PR #4, opened 2024-01-10
Add brevo contacts api

RainbowBunchie:
This contains the basic brevo contacts api. I will implement the missing functionality (marked as `TODO` in the code) in followup MRs. 

This is part of COM-267


.env
```diff
@@ -67,3 +67,6 @@ NEXT_PUBLIC_SITE_IS_PREVIEW=false
 JAEGER_UI_PORT=16686
 JAEGER_OLTP_PORT=4318
 TRACING_ENABLED=1
+
+# BREVO
+BREVO_API_KEY="PLEASE OVERWRITE IN .env.local FILE"
```
raphaelblum, line 72: Couldn't you remove the variable?  Then the API Config validation should throw a validation error that the key needs to be set.

And I would say we describe in the README what needs to be done to start the project (setting the BREVO_API_KEY in the .env.local file)

packages/api/src/brevo-contact/brevo-contact.module.ts
```diff
@@ -0,0 +1,12 @@
+import { MikroOrmModule } from "@mikro-orm/nestjs";
+import { Module } from "@nestjs/common";
+
+import { BrevoModule } from "../brevo/brevo.module";
+import { BrevoContactResolver } from "./brevo-contact.resolver";
+import { BrevoContactsService } from "./brevo-contacts.service";
+
+@Module({
+    imports: [MikroOrmModule.forFeature([]), BrevoModule],
```
raphaelblum, line 9: `MikroOrmModule.forFeature([]),`
I think that shouldn't be necessary

packages/api/src/brevo-contact/brevo-contact.resolver.ts
```diff
@@ -0,0 +1,41 @@
...
+import { BrevoContact } from "./dto/brevo-contact";
+import { BrevoContactUpdateInput } from "./dto/brevo-contact.input";
+import { BrevoContactsArgs } from "./dto/brevo-contacts.args";
+import { PaginatedBrevoContacts } from "./dto/paginated-brevo-contact";
+
+@Resolver(() => BrevoContact)
+export class BrevoContactResolver {
+    constructor(
+        private readonly brevoContactsService: BrevoContactsService,
+        private readonly brevoContactApiService: BrevoContactsApiService, // @Inject(CONFIG) private readonly config: Config,
```
raphaelblum, line 14: // @Inject(CONFIG) private readonly config: Config,
the comment can be removed

RainbowBunchie: Removed in [a04b768](https://github.com/vivid-planet/comet-brevo-module/pull/4/commits/a04b768087c44aa5533f33cb84789b64bae9367d)

packages/api/src/brevo/brevo.utils.ts
```diff
@@ -0,0 +1,5 @@
+import * as http from "http";
+
+export function isErrorFromBrevo(error: unknown): error is { response: http.IncomingMessage } {
+    return typeof error === "object" && error !== null && "response" in error && error.response instanceof http.IncomingMessage;
+}
```
raphaelblum, line 5: Usually we say that utils functions outside a service are not nest-like. What is your opinion in such a case, @johnnyomair ?

I think we should create a utils service? Maybe "brevo-api-utils.service.ts"

And then there is a public function in this service which we can use in the brevo-contacts-api.service.ts, brevo-campaigns-api.service.ts and so on

raphaelblum: Okay, then it's fine for me like it is currently.

We have to improve the error handling for Brevo in the future anyway, I'll make a note of that.

raphaelblum (dismissed):

demo/api/src/app.module.ts
```diff
@@ -125,7 +125,13 @@ export class AppModule {
                 ProductsModule,
                 MenusModule,
                 DependenciesModule,
-                BrevoModule,
+                BrevoModule.register({
+                    api: {
+                        brevo: {
```
johnnyomair, line 130: These two levels in the config seem unnecessary to me. Why did you add them?

packages/api/src/brevo-contact/brevo-contact.module.ts
```diff
@@ -0,0 +1,11 @@
...
+
+import { BrevoModule } from "../brevo/brevo.module";
+import { BrevoContactResolver } from "./brevo-contact.resolver";
+import { BrevoContactsService } from "./brevo-contacts.service";
+
+@Module({
+    imports: [BrevoModule],
+    providers: [BrevoContactsService, BrevoContactResolver],
+})
+export class BrevoContactModule {}
```
johnnyomair, line 11: Why is this not part of the brevo module? What are you thoughts behind this decision?

packages/api/src/brevo-module.ts
```diff
@@ -1,11 +1,16 @@
...
+import { ConfigModule } from "./config/config.module";
+
 @Global()
 @Module({})
 export class BrevoModule {
-    static register(): DynamicModule {
+    static register(config: BrevoModuleConfig): DynamicModule {
         return {
             module: BrevoModule,
+            imports: [BrevoModule, BrevoContactModule, ConfigModule.forRoot(config)],
```
johnnyomair, line 13: The BrevoModule can't import itself?

packages/api/src/config/config.module.ts
```diff
@@ -0,0 +1,21 @@
+import { DynamicModule, Global, Module } from "@nestjs/common";
+
+import { BrevoModuleConfig } from "./brevo-module.config";
+import { BREVO_MODULE_CONFIG } from "./brevo-module.constants";
+
+@Global()
+@Module({})
+export class ConfigModule {
+    static forRoot(config: BrevoModuleConfig): DynamicModule {
```
johnnyomair, line 9: IMO there's no need for this to be a separate module. The Brevo module should provide the config

packages/api/src/db/ormconfig.cli.ts
```diff
@@ -0,0 +1,11 @@
+import { Options } from "@mikro-orm/core";
+
+import { ormConfig } from "./ormconfig";
+
+const config: Options = {
```
johnnyomair, line 5: The package shouldn't have this file, this should be part of the demo

packages/api/src/db/ormconfig.ts
```diff
@@ -0,0 +1,28 @@
+import { createMigrationsList, createOrmConfig } from "@comet/cms-api";
+import { EntityCaseNamingStrategy } from "@mikro-orm/core";
+import path from "path";
+
+export const ormConfig = createOrmConfig({
```
johnnyomair, line 5: Here as well

packages/api/package.json
```diff
@@ -23,23 +24,54 @@
...
-        "typescript": "^4.0.0"
+        "typescript": "^4.0.0",
+        "uuid": "^8.3.2"
+    },
+    "mikro-orm": {
+        "useTsNode": true,
+        "configPaths": [
+            "./src/db/ormconfig.cli.ts",
+            "./dist/db/ormconfig.cli.js"
+        ]
```
johnnyomair, line 75: This isn't needed in the package

packages/api/src/brevo/brevo.module.ts
```diff
@@ -0,0 +1,11 @@
...
+
+import { ConfigModule } from "../config/config.module";
+import { BrevoContactsApiService } from "./brevo-contact-api.service";
+
+@Module({
+    imports: [ConfigModule],
+    providers: [BrevoContactsApiService],
+    exports: [BrevoContactsApiService],
+})
+export class BrevoModule {}
```
raphaelblum, line 11: why are there 2 BrevoModules?

RainbowBunchie: I dropped the 'comet' prefix in the src-folders root module.. I did not realize that the naming now exists 2 times. I guess this should now rather be called `BrevoApiModule` since this module contains all brevo logic. 
I will change the naming!

demo/api/src/app.module.ts
RainbowBunchie: I thought I could use a global config for both admin and api in the beginning that's why I nested it like this. But since admin will be a seperate package and the config cannot be used for both this is not needed. I removed the unnecessary api nesting in f3fbde7e83b8c5d0000bc1c014d19e7db94e308d

packages/api/src/brevo-contact/brevo-contact.module.ts
RainbowBunchie: Also the brevoModule will contain the brevo api logic for the campaigns & target groups in the future as well.

packages/api/src/brevo-module.ts
RainbowBunchie: The package contained 2 BrevoModules due to the previous drop of the `comet` prefix of the root module. I renamed one of the modules to 'BrevoApiModule` which contains all the logic for the communication with brevo. Changed in 4d9b89ebf57c92cdc8217a5bb77b5a4b03419635

packages/api/src/db/ormconfig.ts
RainbowBunchie: I used the api generator to generate the basic files for the 'BrevoContacts' and the api generator needs the ormconfig to be present . Or is there another way to use the api generator? I can remove the orm config after I generated all the files for the future entities.

packages/api/src/brevo/brevo.module.ts
RainbowBunchie: Changed in 4d9b89ebf57c92cdc8217a5bb77b5a4b03419635

packages/api/src/config/config.module.ts
RainbowBunchie: I wanted to extract this into a seperate config module so it can easily imported in further modules as well (target groups, campaigns,..). Else the config must be manually passed to each module in the (root) BrevoModule and. I think this keeps the modules a bit cleaner. Do you still think it would be better to not have the separate module?

packages/api/src/db/ormconfig.ts
johnnyomair: You'll probably need to run the generator in the Demo project and then copy the files to the library. We also do this for migrations: We generate them in Demo and then move it to the library package.

@nsams do you have a better idea for using the generator in a library package?

demo/api/src/app.module.ts
```diff
@@ -125,7 +125,11 @@ export class AppModule {
                 ProductsModule,
                 MenusModule,
                 DependenciesModule,
-                BrevoModule,
+                BrevoModule.register({
+                    brevo: {
```
johnnyomair, line 129: I'd also remove the brevo level, since it's clear that this is the config for Brevo. Will there ever be another key in the config?

RainbowBunchie: Yes there will be multiple other keys that are not related to brevo directly (eg Scope, ContentBlock, ..) that's why I kept the brevo level here.

johnnyomair (approved):

packages/api/src/db/ormconfig.ts
RainbowBunchie: @johnnyomair I removed the files for now, so this does not block this MR. When there is a better solution for this I will adapt it in the future and I will stick to generating in the demo for now. Is this alright with you?

see [fffcb10](https://github.com/vivid-planet/comet-brevo-module/pull/4/commits/fffcb10e3f52e1047598bbae71f6dac9281ab4d5)

johnnyomair (approved):

raphaelblum (approved):

PR #5, opened 2024-01-12
Add ecg rtr list for subscription of contact

RainbowBunchie:
depends on #4 

Add ecg rtr list check to email addresses when a contact wants to subscribe to a list 

Part of COM-267


RainbowBunchie:
Moved to #7

PR #11, opened 2024-01-18
add changeset and pull request template

raphaelblum:


.changeset/config.json
```diff
@@ -0,0 +1,14 @@
+{
+    "$schema": "https://unpkg.com/@changesets/config@2.3.0/schema.json",
+    "changelog": "@changesets/cli/changelog",
+    "commit": false,
+    "fixed": [["@comet/*"]],
```
raphaelblum, line 5: I'm not sure if I need that but I guess yes?
https://github.com/changesets/changesets/blob/main/docs/config-file-options.md#fixed-array-of-arrays-of-package-names

I guess the comet brevo module should be published like that (with a leading `@`):
 -  comet/brevo-module-api
 -  comet/brevo-module-admin

johnnyomair: Yes, you'll need that to have same versions for both packages 👍🏼

package.json
```diff
@@ -28,5 +28,8 @@
         "prettier": "^2.3.2",
         "ts-node": "^10.5.0",
         "typescript": "^4.2.2"
+    },
+    "dependencies": {
+        "@changesets/cli": "^2.27.1"
```
johnnyomair, line 33: Please move this to dev dependencies (this is kinda wrong in Comet)

raphaelblum: done in [bc54768d90c671562114c5067325b4e74b04973e](https://github.com/vivid-planet/comet-brevo-module/pull/11/commits/bc54768d90c671562114c5067325b4e74b04973e)

johnnyomair (approved):

PR #16, opened 2024-01-22
Add brevo campaigns api to email campaign

RainbowBunchie:
Add brevo api functionality to email campaigns

This is part of COM-266

### DIFF
--- a/packages/api/package.json
+++ b/packages/api/package.json
@@ -23,6 +23,8 @@
         "test:watch": "jest --watch"
     },
     "dependencies": {
+        "@nestjs/axios": "^1.0.0",
+        "@sendinblue/client": "^3.3.1",
         "node-fetch": "^2.6.1"
     },
     "devDependencies": {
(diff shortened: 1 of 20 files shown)


packages/api/package.json
```diff
@@ -31,10 +31,13 @@
         "@mikro-orm/nestjs": "^5.0.0",
         "@mikro-orm/postgresql": "^5.0.4",
         "@nestjs/common": "^9.0.0",
+        "axios": "^0.21.0",
         "reflect-metadata": "^0.1.13",
         "rxjs": "^7.0.0"
     },
     "dependencies": {
+        "@nestjs/axios": "^1.0.0",
+        "@sendinblue/client": "^3.3.1",
```
raphaelblum, line 40: i think this should be a peer dependency of the brevo module, shouldn't it?

RainbowBunchie: Since this is used only in the module and not in any of the comet core packages I think this needs to be a 'normal' dependency. But I am still quite unsure about the dependency sections.

raphaelblum (approved):

PR #20, opened 2024-01-23
COM-272: Add brevo contacts admin

RainbowBunchie:
I used https://github.com/vivid-planet/comet/blob/main/packages/admin/cms-admin/src/redirects/createRedirectsPage.tsx as a reference for the custom scope. 

The peer packages are all needed to to the import of `@comet/cms-admin` which is needed for the scope.

Configurable attributes are still missing in the grid (marked as `TODO` in the code). Right now only the fixed properties are shown.
Screenshot 2024-01-23 at 13 38 30
Screenshot 2024-01-23 at 13 38 36

This is part of COM-272


packages/admin/src/brevoContacts/BrevoContactsGrid.tsx
```diff
@@ -0,0 +1,202 @@
...
+
+export function BrevoContactsGrid({ scope }: { scope: ContentScopeInterface }): React.ReactElement {
+    const client = useApolloClient();
+    const intl = useIntl();
+    const dataGridProps = { ...useDataGridRemote(), ...usePersistentColumnState("BrevoContactsGrid") };
+
+    const columns: GridColDef<GQLBrevoContactsListFragment>[] = [
+        {
+            field: "createdAt",
+            headerName: intl.formatMessage({ id: "comet.brevoContact.subscribedAt", defaultMessage: "Subscribed At" }),
```
raphaelblum, line 92: usually the comet prefix is reserved for comet directly. Since this is a comet module, it could be okay but I'd still suggest another prefix. At some point we could define a key in the same way as in main comet library, and then we would have the same key twice with a different default message, which could lead to confusion at some point (because afaik it would overwrite one defaultMessage then)

So I'd suggest brevoModule.brevoContact, or do you have a better naming suggestion?

RainbowBunchie: I changed it to `cometBrevoModule` in [853a758](https://github.com/vivid-planet/comet-brevo-module/pull/20/commits/853a75879e572e8e9e623d2ee1c4f26455666410)

RainbowBunchie:
@raphaelblum I found a small bug which displayed the wrong action menu label/icon for the block/unblock logic 
-> fixed in [2befbe9](https://github.com/vivid-planet/comet-brevo-module/pull/20/commits/2befbe940b7d03873b9180906b60faded639d1af)

raphaelblum (approved):

PR #30, opened 2024-01-26
COM-271: Add dynamic fields to target group form

RainbowBunchie:
depends on #29 

This adds configurable fields to the target group form 

With configuration (code in this PR):
Screenshot 2024-01-26 at 11 20 42

Without configuration the filters section will not be rendered:
Screenshot 2024-01-26 at 11 22 13

This is part of COM-271

### DIFF
--- a/packages/admin/src/targetGroups/TargetGroupForm.tsx
+++ b/packages/admin/src/targetGroups/TargetGroupForm.tsx
@@ -44,10 +44,10 @@ interface FormProps {
     scope: ContentScopeInterface;
     additionalFormFields?: React.ReactNode;
     nodeFragment?: { name: string; fragment: DocumentNode };
-    dataToInitialValues?: (values?: EditTargetGroupFinalFormValues) => EditTargetGroupFinalFormValues;
+    input2State?: (values?: EditTargetGroupFinalFormValues) => EditTargetGroupFinalFormValues;
 }
 
-export function TargetGroupForm({ id, scope, additionalFormFields, dataToInitialValues, nodeFragment }: FormProps): React.ReactElement {
+export function TargetGroupForm({ id, scope, additionalFormFields, input2State, nodeFragment }: FormProps): React.ReactElement {
     const stackApi = useStackApi();
     const client = useApolloClient();
     const mode = id ? "edit" : "add";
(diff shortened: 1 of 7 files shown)


packages/admin/src/targetGroups/TargetGroupForm.tsx
```diff
@@ -27,41 +28,54 @@ import { createTargetGroupMutation, targetGroupFormQuery, updateTargetGroupMutat
...
-        [data],
-    );
+    const initialValues = React.useMemo<Partial<EditTargetGroupFinalFormValues>>(() => {
+        let additionalInitialValues = {};
+
+        if (dataToInitialValues) {
+            additionalInitialValues = dataToInitialValues(data?.targetGroup);
+        }
+
+        return data?.targetGroup ? { title: data.targetGroup.title, ...additionalInitialValues } : { ...additionalInitialValues };
```
raphaelblum, line 77:
```suggestion
        return data?.targetGroup ? { title: data.targetGroup.title, ...additionalInitialValues } : additionalInitialValues ;
```

packages/admin/src/targetGroups/TargetGroupForm.tsx
```diff
@@ -27,41 +28,54 @@ import { createTargetGroupMutation, targetGroupFormQuery, updateTargetGroupMutat
...
+    title: string;
+    [key: string]: unknown;
+}
 
 interface FormProps {
     id?: string;
     scope: ContentScopeInterface;
+    additionalFormFields?: React.ReactNode;
+    nodeFragment?: { name: string; fragment: DocumentNode };
+    dataToInitialValues?: (values?: EditTargetGroupFinalFormValues) => EditTargetGroupFinalFormValues;
```
raphaelblum, line 47: I think I would call it input2State, would be more "comet"-like and blocks also always have the naming: input2State, output2State, state2output

or here for example: https://github.com/vivid-planet/comet/blob/main/packages/admin/cms-admin/src/common/useEditState.tsx

RainbowBunchie: I renamed the prop to `input2State` in [c23831d](https://github.com/vivid-planet/comet-brevo-module/pull/30/commits/c23831d30073b719fdca309fb65cc8b5c48c030e)

packages/admin/src/targetGroups/TargetGroupForm.tsx
RainbowBunchie: adapted in [8cdbcf4](https://github.com/vivid-planet/comet-brevo-module/pull/30/commits/8cdbcf48df375f6a0b5a8125fe8fe9f594da1201)

raphaelblum (approved):

PR #32, opened 2024-02-01
COM-343: Add campaigns admin

RainbowBunchie:
Add admin grid & form for email campaigns. 
I will add a section for test emails & a "send now" button in a follow up PR. 
The preview in the demo is not working since there is no frontend for the email campaigns as of now in the demo. 

Screenshot 2024-02-01 at 09 01 44
Screenshot 2024-02-01 at 09 02 05
Screenshot 2024-02-01 at 09 02 18
Screenshot 2024-02-01 at 09 02 43

This is part of COM-343

PR #33, opened 2024-02-02
COM-343: Add additional functionality to email campaign admin (send now & test emails)

RainbowBunchie:
depends on #32 

Screenshot 2024-02-02 at 13 19 20

This is part of COM-343

### DIFF
--- a/packages/api/src/email-campaign/email-campaigns.service.ts
+++ b/packages/api/src/email-campaign/email-campaigns.service.ts
@@ -125,6 +125,6 @@ export class EmailCampaignsService {
             } while (currentOffset < totalContacts);
         }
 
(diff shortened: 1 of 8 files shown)


packages/api/src/email-campaign/email-campaigns.service.ts
```diff
@@ -125,6 +125,6 @@ export class EmailCampaignsService {
             } while (currentOffset < totalContacts);
         }
```
raphaelblum, line 126: not inthis PR:

but as discussed, this function is wrong. the sendBrevoCampaign has to be put out of the while loop.

raphaelblum (approved):

PR #36, opened 2024-02-07
fix validation error when saving campaigns: IsUndefined validator correctly throwed an error when sending null

raphaelblum:
the error probably occurred when we changed the IsNullable to IsUndefinable in the api: https://github.com/vivid-planet/comet-brevo-module/pull/32#discussion_r1476001379

### DIFF
--- a/packages/admin/src/emailCampaigns/form/EmailCampaignForm.tsx
+++ b/packages/admin/src/emailCampaigns/form/EmailCampaignForm.tsx
@@ -100,7 +100,7 @@ export function EmailCampaignForm({ id, EmailCampaignContentBlock, scope, previe
                 title: emailCampaign?.title,
                 subject: emailCampaign?.subject,
                 content: EmailCampaignContentBlock.input2State(emailCampaign.content),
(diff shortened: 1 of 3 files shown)


johnnyomair (dismissed):

johnnyomair:
You could also consider adding `@IsNullable()` to the mutation, depending on the use case... Is it possible to set the field to `null`?

raphaelblum:
> You could also consider adding `@IsNullable()` to the mutation, depending on the use case... Is it possible to set the field to `null`?

I hope I have understood you correctly. I have now done it like this: if you send undefined, the previously set value remains. if you send null, the previously set value is removed (for scheduledAt)

couldn't I just use @IsOptional if I support undefined and null?

johnnyomair:
> I hope I have understood you correctly. I have now done it like this: if you send undefined, the previously set value remains. if you send null, the previously set value is removed (for scheduledAt)

Yes, you understood that correctly.

> couldn't I just use @IsOptional if I support undefined and null?

Yes. IMO the other two decorators are more descriptive, but that's personal preference.

johnnyomair (approved):

RainbowBunchie (approved):

PR #40, opened 2024-05-06
Update to Comet v6

stekalt:
# **Breaking Changes**:

- Now requires >= v6.0.0 for `@comet` packages

## API:

- All GraphQL resolvers now require the `brevo-newsletter` permission.
- `BrevoContactResolver#subscribeBrevoContact` mutation: The `scope` argument was moved outside `input` to enable an automatic scope check
- `BrevoContactsService#createDoubleOptInContact`: `scope` was moved outside `data` and is now the second argument
- `TargetGroupsService#findNonMainTargetGroups`: `data` was replaced with `scope`

---

COM-708


johnnyomair:
@thomasdax98 should I review this PR or @raphaelblum?

.changeset/shaggy-feet-live.md
```diff
@@ -0,0 +1,9 @@
+---
+"@comet/brevo-admin": major
+---
+
+Make this package compatible with [COMET v6](https://docs.comet-dxp.com/docs/migration/migration-from-v5-to-v6)
+
+**Breaking Changes**:
+
+- Now requires >= v6.0.0 for `@comet` packages
```
raphaelblum, line 9: The file can be deleted, right?
Everything is in the other changeset file as well

.env
```diff
@@ -69,4 +69,5 @@ JAEGER_OLTP_PORT=4318
 TRACING_ENABLED=1
 
 # BREVO
-BREVO_ALLOWED_REDIRECT_URL=http://${DEV_DOMAIN:-localhost}${WORKAROUND_DOTENV_ISSUE}:${SITE_PORT}
\ No newline at end of file
+BREVO_ALLOWED_REDIRECT_URL=http://${DEV_DOMAIN:-localhost}${WORKAROUND_DOTENV_ISSUE}:${SITE_PORT}
+CAMPAIGNS_FRONTEND_URL=http://${DEV_DOMAIN:-localhost}${WORKAROUND_DOTENV_ISSUE}:${SITE_PORT} # doesn't work, TODO: add actual mailing frontend
```
raphaelblum, line 73: FYI
@RainbowBunchie added this here as well:
https://github.com/vivid-planet/comet-brevo-module/pull/41

packages/api/.eslintrc.json
```diff
@@ -1,4 +1,7 @@
 {
     "extends": "@comet/eslint-config/nestjs",
-    "ignorePatterns": ["lib/**"]
+    "ignorePatterns": ["lib/**"],
+    "rules": {
+        "@comet/no-other-module-relative-import": "off"
+    }
```
raphaelblum, line 6: why is this off?

i think we should enable that, maybe in an extra PR if it is a lot work.

.env
RainbowBunchie: @raphaelblum 
I rebased onto this branch - so we will not get any problems with this env and also with the comet versions

.changeset/shaggy-feet-live.md
thomasdax98: This changeset is for `brevo-admin`, the other is for `brevo-api`. Both only contain the relevant info for this package, so I'd keep both

The alternative would be to share one changeset for both packages. But then the changeset is copied twice to the resulting changelog and the `brevo-admin` section contains irrelevant info.

.env
thomasdax98: Thanks @RainbowBunchie

packages/api/.eslintrc.json
thomasdax98: I copied this from the comet repo where we disabled it for all packages (see https://github.com/vivid-planet/comet/pull/1303/commits/2164c97620addd8d40f97d87c593427629b03cf7).

But I'm not sure what the reason was. If I remember correctly, there was some problem with setting up the `@src` alias in a package. Maybe @johnnyomair can comment?

.changeset/shaggy-feet-live.md
raphaelblum: >  This changeset is for brevo-admin, the other is for brevo-api

👍

packages/api/.eslintrc.json
johnnyomair: We can't have an import alias in the packages because that's a feature from transpilers such as Babel. TSC doesn't support it directly, and we decided to not use a workaround that patches it to achieve the import alias.

raphaelblum (approved):

demo/api/src/products/product.crud.resolver.ts
```diff
@@ -1,28 +1,26 @@
 // Scaffolded by the CRUD generator on 2023-03-20.
```
johnnyomair, line 1: Out of scope: Should we remove products altogether? Is this needed for the module?

packages/api/src/brevo-contact/dto/brevo-contacts.args.ts
```diff
@@ -23,7 +23,6 @@ export class BrevoContactsArgsFactory {
 
             @Field(() => String, { nullable: true })
             @IsOptional()
-            @IsEmail()
```
johnnyomair, line 26: Why do you remove this?

thomasdax98: Because the email field for this query is filled by a search input. This causes an error when typing part of an email address. 

For example, I want to search for `test@example.com` but typing `t` already causes a validation error

demo/api/src/products/product.crud.resolver.ts
thomasdax98: I don't think it's needed (@raphaelblum, @RainbowBunchie ?). So yes, we could remove it completely. But that's not urgent

johnnyomair (approved):

demo/api/src/products/product.crud.resolver.ts
RainbowBunchie: They are not needed, so they can be removed in the future

PR #41, opened 2024-05-13
Add demo mailing frontend

RainbowBunchie:
depends on #40 

This adds an email campaign frontend for the email campaign preview and also for the generation of the email content that gets send to brevo in the demo. 

image

COM-391

PR #42, opened 2024-05-21
#41 View sent campaigns

RainbowBunchie:
depends on #41

Screenshot 2024-05-21 at 09 01 40
Screenshot 2024-05-21 at 09 01 49

### DIFF
--- a/packages/admin/src/emailCampaigns/form/EmailCampaignForm.tsx
+++ b/packages/admin/src/emailCampaigns/form/EmailCampaignForm.tsx
@@ -20,11 +20,9 @@ import {
     BlocksFinalForm,
     BlockState,
     createFinalFormBlock,
-    IFrameBridgeProvider,
     parallelAsyncEvery,
 } from "@comet/blocks-admin";
 import {
-    BlockPreview,
     BlockPreviewWithTabs,
     ContentScopeInterface,
     EditPageLayout,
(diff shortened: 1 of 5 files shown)


packages/admin/src/emailCampaigns/form/EmailCampaignForm.tsx
```diff
@@ -227,77 +230,85 @@ export function EmailCampaignForm({ id, EmailCampaignContentBlock, scope, previe
...
-                            ),
-                        },
-                    ]}
-                </BlockPreviewWithTabs>
+                {isViewOnly ? (
+                    <IFrameBridgeProvider key={previewUrl}>
+                        <BlockPreview url={previewUrl} previewState={previewState} previewApi={previewApi} />
+                    </IFrameBridgeProvider>
+                ) : (
+                    <BlockPreviewWithTabs previewUrl={previewUrl} previewState={previewState} previewApi={previewApi}>
```
raphaelblum, line 241: Optional:
I think it might be better to split the component into two components, because with a few more functions this EmailCampaignForm will become quite complex, but it would be okay for me to leave it as it is in the first step.

raphaelblum (dismissed):

packages/admin/src/emailCampaigns/form/EmailCampaignForm.tsx
RainbowBunchie: 6fb2f17bc8f4e132c8d42ebb0422e6e4f2e2b4ce

raphaelblum (approved):

PR #46, opened 2024-06-10
Edit brevo contact

RainbowBunchie:
-   [x] Add changeset (if necessary)

Add an edit page for the contact 
Screenshot 2024-06-10 at 13 56 23
Screenshot 2024-06-10 at 13 57 06

PR #47, opened 2024-06-11
Add brevo contact

RainbowBunchie:
depends on #46 

Screenshot 2024-06-11 at 10 21 38
Screenshot 2024-06-11 at 10 22 21


demo/admin/src/common/brevoModuleConfig/brevoContactsPageAttributesConfig.tsx
```diff
@@ -112,6 +113,7 @@ export const getBrevoContactConfig = (
         input2State: (values?: AdditionalFormConfigInputProps) => {
             return {
                 email: values?.email ?? "",
+                redirectionUrl: values?.redirectionUrl ?? "",
```
raphaelblum, line 116: shouldn't that be required?

or in this case throw an error if it is not set?

packages/admin/src/brevoContacts/form/BrevoContactForm.tsx
```diff
@@ -174,10 +179,17 @@ export function BrevoContactForm({ id, scope, input2State, additionalFormFields,
...
-                                />
+                                {mode === "edit" ? (
+                                    <FormattedMessage
+                                        id="cometBrevoModule.brevoContact.contactEditAlert"
+                                        defaultMessage="Editing a contact will affect all scopes and the target groups within those scopes."
+                                    />
+                                ) : (
+                                    <FormattedMessage
+                                        id="cometBrevoModule.brevoContact.contactAddAlert"
+                                        defaultMessage="The contact will get a DOI email to confirm the subscription. After the contact's confirmation, the contact will be added to the corresponding target groups in this scope depending on the contact's attributes. Before the confirmation the contact will not be shown on the contacts page."
```
raphaelblum, line 190: I would write DOI out:
"Double Opt-In" or "DOI(Double Opt-In)"

Not all users will know what Double Opt-In is and even less will know the abbreviation DOI.
At least with Double Opt-In you can easily google what it is.

demo/admin/src/common/brevoModuleConfig/brevoContactsPageAttributesConfig.tsx
RainbowBunchie: It is required on create. But not on update. Both in api and admin.
Please check `packages/api/src/brevo-contact/dto/brevo-contact-input.factory.ts` and `packages/admin/src/brevoContacts/form/BrevoContactForm.tsx`

packages/admin/src/brevoContacts/form/BrevoContactForm.tsx
RainbowBunchie: You are right. I change `DOI` to `double opt-in` in lower case. Since brevo is using the lowercase version in their explanation. https://www.brevo.com/blog/double-opt-in/

RainbowBunchie: [ec28a03](https://github.com/vivid-planet/comet-brevo-module/pull/47/commits/ec28a03d296ef5af7b586a9b977b3aa9c453484c)

raphaelblum (approved):

PR #49, opened 2024-06-14
Add further config fields per scope

RainbowBunchie:

TODO: 

- [ ] set correct brevo api key per scope
- [ ] remove api key from brevo config in code + env var

### DIFF
--- a/packages/admin/src/brevoConfiguration/BrevoConfigForm.tsx
+++ b/packages/admin/src/brevoConfiguration/BrevoConfigForm.tsx
@@ -3,6 +3,7 @@ import {
     Field,
     FinalForm,
     FinalFormAutocomplete,
+    FinalFormInput,
     FinalFormSaveSplitButton,
     FinalFormSubmitEvent,
     Loading,
@@ -19,12 +20,20 @@ import { FormApi } from "final-form";
 import React from "react";
 import { FormattedMessage } from "react-intl";
 
-import { brevoConfigFormQuery, createBrevoConfigMutation, sendersSelectQuery, updateBrevoConfigMutation } from "./BrevoConfigForm.gql";
+import {
+    brevoConfigFormQuery,
+    createBrevoConfigMutation,
+    doiTemplatesSelectQuery,
+    sendersSelectQuery,
+    updateBrevoConfigMutation,
+} from "./BrevoConfigForm.gql";
 import {
     GQLBrevoConfigFormQuery,
     GQLBrevoConfigFormQueryVariables,
     GQLCreateBrevoConfigMutation,
     GQLCreateBrevoConfigMutationVariables,
+    GQLDoiTemplatesSelectQuery,
+    GQLDoiTemplatesSelectQueryVariables,
     GQLSendersSelectQuery,
     GQLSendersSelectQueryVariables,
     GQLUpdateBrevoConfigMutation,
@@ -36,7 +45,9 @@ interface Option {
     label: string;
 }
 type FormValues = {
-    sender: Option;
+    sender?: Option;
+    apiKey?: string;
+    doiTemplate?: Option;
 };
 
 interface FormProps {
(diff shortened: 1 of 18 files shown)


.changeset/sweet-plants-nail.md
```diff
@@ -0,0 +1,23 @@
...
+```diff
+BrevoModule.register({
+    brevo: {
+-       apiKey: config.brevo.apiKey,
+-       doubleOptInTemplateId: config.brevo.doubleOptInTemplateId,
+        //...
+    },
+    //..
+})
+```
```
thomasdax98, line 23: Adding the scope to most resolvers and services is a breaking change. I think there should at least be a major changeset mentioning which this. It doesn't have to be detailed since we know it's not relevant in detail

packages/api/src/brevo-api/brevo-api-campaigns.service.ts
```diff
@@ -12,9 +16,26 @@ import { BrevoApiCampaignStatistics } from "./dto/brevo-api-campaign-statistics"
 export class BrevoApiCampaignsService {
     private readonly campaignsApi: SibApiV3Sdk.EmailCampaignsApi;
 
-    constructor(@Inject(BREVO_MODULE_CONFIG) private readonly config: BrevoModuleConfig) {
+    constructor(
+        @Inject(BREVO_MODULE_CONFIG) private readonly config: BrevoModuleConfig,
+        @InjectRepository("BrevoConfig") private readonly brevoConfigRepository: EntityRepository<BrevoConfigInterface>,
```
thomasdax98, line 21: I think `"BrevoConfig"` should be a constant defined in `brevo-config-entity.factory.ts`

packages/api/src/brevo-config/entities/brevo-config-entity.factory.ts
```diff
@@ -23,19 +26,29 @@ export class BrevoConfigEntityFactory {
...
+
+            @Property({ columnType: "number", nullable: true })
+            @Field(() => Int, { nullable: true })
+            doiTemplateId?: number;
+
+            @Property({ columnType: "text", nullable: true })
+            apiKey?: string;
+
+            @Field(() => Boolean)
+            isApiKeySet: boolean;
```
thomasdax98, line 51: I think this can be removed since it's determined in the field resolver and isn't actually set here

packages/admin/src/common/configVerification/ConfigVerification.tsx
```diff
@@ -0,0 +1,40 @@
...
+import { GQLBrevoConfigCheckQuery, GQLBrevoConfigCheckQueryVariables } from "./ConfigVerification.gql.generated";
+
+interface ConfigCheckProps {
+    scope: ContentScopeInterface;
+}
+
+export function ConfigVerification({ scope, children }: React.PropsWithChildren<ConfigCheckProps>): JSX.Element {
+    const { data, error, loading } = useQuery<GQLBrevoConfigCheckQuery, GQLBrevoConfigCheckQueryVariables>(brevoConfigCheckQuery, {
+        variables: { scope },
+        fetchPolicy: "network-only",
```
thomasdax98, line 17: Can we change this to `cache-and-network` for a faster response with slow connection if the api key is configured and was already loaded on another page?

packages/admin/src/brevoConfiguration/BrevoConfigForm.tsx
```diff
@@ -106,47 +159,67 @@ export function BrevoConfigForm({ scope }: FormProps): React.ReactElement {
...
 
-    if (error || senderError) throw error ?? senderError;
+    if (error || senderError || doiTemplatesError) throw error ?? senderError ?? doiTemplatesError;
 
-    if (loading || senderLoading) {
+    if (loading || senderLoading || doiTemplatesLoading) {
         return <Loading behavior="fillPageHeight" />;
     }
 
+    const isApiKeySet = mode !== "add" && data?.brevoConfig?.isApiKeySet;
```
thomasdax98, line 221: You could define this helper higher up and use it everywhere (instead of using data?.brevoConfig?.isApiKeySet directly)

PR #53, opened 2024-06-20
Add brevo config via code

RainbowBunchie:


packages/api/src/brevo-api/brevo-api-campaigns.service.ts
```diff
@@ -14,7 +15,11 @@ export class BrevoApiCampaignsService {
 
     constructor(@Inject(BREVO_MODULE_CONFIG) private readonly config: BrevoModuleConfig) {
```
thomasdax98, line 16: Not directly related to this PR but while we are at it: Why is `BREVO_MODULE_CONFIG` not provided by the `BrevoModule`? That would make more sense to me. It doesn't seem completely logical that I have to provide the config via the `ConfigModule` AND pass it to the `BrevoModule`

packages/api/src/brevo-api/brevo-api-campaigns.service.ts
```diff
@@ -14,7 +15,11 @@ export class BrevoApiCampaignsService {
...
         this.campaignsApi = new SibApiV3Sdk.EmailCampaignsApi();
-        this.campaignsApi.setApiKey(SibApiV3Sdk.EmailCampaignsApiApiKeys.apiKey, config.brevo.apiKey);
+    }
+
+    private setApiKey(scope: EmailCampaignScopeInterface): void {
+        const { apiKey } = this.config.brevo.getBrevoConfig(scope);
```
thomasdax98, line 21: I can't see why this method is available here since it's not defined in the `ConfigModule` (but only in `BrevoModule.register`). I guess I'm missing something here(?)

thomasdax98 (dismissed):

demo/api/src/db/migrations/Migration20240619145217.ts
```diff
@@ -0,0 +1,7 @@
+import { Migration } from "@mikro-orm/migrations";
+
+export class Migration20240619145217 extends Migration {
```
johnnyomair, line 3: I believe this migration should be in the package,

packages/api/schema.gql
```diff
@@ -173,29 +159,18 @@ type PaginatedEmailCampaigns {
...
-  scope: EmailCampaignContentScope!
-}
-
 input EmailCampaignContentScopeInput {
   thisScopeHasNoFields____: String
 }
 
 type Query {
-  brevoContact(id: Int!): BrevoContact!
+  brevoContact(id: Int!, scope: EmailCampaignContentScopeInput!): BrevoContact!
```
johnnyomair, line 167: Is it necessary to pass the scope here? Shouldn't the ID be unique across scopes?

packages/api/schema.gql
```diff
@@ -263,9 +238,9 @@ enum EmailCampaignSortField {
 }
 
 type Mutation {
-  updateBrevoContact(id: Int!, input: BrevoContactUpdateInput!): BrevoContact!
+  updateBrevoContact(id: Int!, scope: EmailCampaignContentScopeInput!, input: BrevoContactUpdateInput!): BrevoContact!
```
johnnyomair, line 241: Here as well

packages/api/schema.gql
```diff
@@ -263,9 +238,9 @@ enum EmailCampaignSortField {
...
   createBrevoContact(scope: EmailCampaignContentScopeInput!, input: BrevoContactInput!): SubscribeResponse!
-  deleteBrevoContact(id: Int!): Boolean!
+  deleteBrevoContact(id: Int!, scope: EmailCampaignContentScopeInput!): Boolean!
```
johnnyomair, line 243: Here as well

packages/api/src/brevo-api/brevo-api-campaigns.service.ts
```diff
@@ -14,7 +15,11 @@ export class BrevoApiCampaignsService {
...
+        this.campaignsApi.setApiKey(SibApiV3Sdk.EmailCampaignsApiApiKeys.apiKey, apiKey);
```
johnnyomair, line 22: I believe this isn't a good idea: One could forget to call the the setApiKey-method, and as the service isn't request-scoped, the API key of the previous request will be used. A better approach would be to remove the campaignsApi member and add a `getCampaignsApi(scope)` method that creates a new campaign API for the scope (the result could be cached in a map).

packages/api/src/config/brevo-module.config.ts
```diff
@@ -6,10 +6,16 @@ import { BrevoContactFilterAttributesInterface } from "../types";
 
 export interface BrevoModuleConfig {
     brevo: {
-        apiKey: string;
+        getBrevoConfig: (scope: EmailCampaignScopeInterface) => {
```
johnnyomair, line 9: IMO getConfig is enough since its already in `brevo`. Could also be `resolveConfig`

packages/api/schema.gql
RainbowBunchie: The problem here is that the contact is not stored in the database with a scope, but simply in brevo without any knowledge which scope the contact belongs too. The contacts of a specific scope are stored inside a "main" target list for the scope and the target group information is stored in the database. The scope cannot easily be stored within the brevo contacts (contact can be within more that one list for example and brevo limits additional attributes to really basic datatypes only) - that's why the target groups is used for storing the scope information. And since a contact can be in multiple scopes the scope cannot be retrieved that way.

packages/api/src/config/brevo-module.config.ts
RainbowBunchie: you are right, I went with `resolveConfig`. Sounds better 
[af37fd8](https://github.com/vivid-planet/comet-brevo-module/pull/53/commits/af37fd8f6b6250b854378548766c47be5ef8b998)

packages/api/src/brevo-api/brevo-api-campaigns.service.ts
RainbowBunchie: I haven't thought about that, this could be really bad. I changed it to your suggested approach: 
Removing the member: [bea2fd3](https://github.com/vivid-planet/comet-brevo-module/pull/53/commits/bea2fd3715bbd88853b4ce66d8f15c1870a1018c)
Adding a map to store the created apis for the scope: [8b1ecf2](https://github.com/vivid-planet/comet-brevo-module/pull/53/commits/8b1ecf231b75a7340b6bcf76461240cc32887b6d)

johnnyomair (approved):

thomasdax98 (approved):

PR #56, opened 2024-07-16
COM-725 Update brevo package

juliawegmayr:

Replace deprecated `sendinblue package with the new brevo-package: https://github.com/getbrevo/brevo-node

### DIFF
--- a/.changeset/good-tools-fry.md
+++ b/.changeset/good-tools-fry.md
@@ -1,0 +1,17 @@
+---
+"@comet/brevo-api": major
+---
+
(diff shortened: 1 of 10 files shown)


.changeset/good-tools-fry.md
```diff
@@ -0,0 +1,17 @@
+---
+"@comet/brevo-api": major
+---
+
+Remove deprecated `sendinblue` package and replace it with `brevo` (https://github.com/getbrevo/brevo-node)
```
raphaelblum, line 5: I think in this case, we really don't need a changeset, since nothing changed for the application.

raphaelblum (approved):

PR #57, opened 2024-07-16
COM-762: add ImageBlock to demo

juliawegmayr:
Add ImageBlock to demo

Screenshot 2024-07-16 at 17 09 45
Screenshot 2024-07-16 at 17 10 10
Screenshot 2024-07-16 at 17 10 27
Screenshot 2024-07-16 at 17 10 35


raphaelblum:
@jamesricky can you please review the frontend/mjml part

raphaelblum (approved):

demo/campaign/src/common/blocks/CommonImageBlock.tsx
```diff
@@ -0,0 +1,78 @@
...
+    const usedCropArea = cropArea ?? damFile.image.cropArea;
+    const usedAspectRatio = calculateInheritAspectRatio(damFile.image, usedCropArea);
+
+    const imageUrl: string = generateImageUrl(
+        {
+            width: getDamAllowedImageWidth(desktopRenderWidth),
+            src: urlTemplate,
+        },
+        usedAspectRatio,
+    );
```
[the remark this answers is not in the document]
juliawegmayr: @jamesricky I added 1020px to the allowedImageSizes as discussed.

I also tried to cover the case that another image size is provided by checking, which validWidth is approximately twice the width given to avoid that images are not displayed at all. Of course, adding the exact value would still be the better solution.

demo/campaign/src/common/blocks/CommonImageBlock.tsx
```diff
@@ -0,0 +1,78 @@
...
+import { css } from "@src/util/stylesHelper";
+import { theme } from "@src/util/theme";
+import * as React from "react";
+
+import { getDamAllowedImageWidth } from "../helpers/imageBlockHelpers";
+
+interface Props extends React.ComponentProps<typeof MjmlImage> {
+    data: PixelImageBlockData;
+    desktopRenderWidth: number;
+    growToOverflowWidth?: boolean;
```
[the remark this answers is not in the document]
jamesricky: You removed parts of this in 782e4e4e4a08a54c45291d0ef135999362533b8a but it still exists in the props type and the styling also still exists.

demo/api/src/comet-config.json
```diff
@@ -6,7 +6,7 @@
     "dam": {
         "additionalMimeTypes": [],
         "uploadsMaxFileSize": 500,
-        "allowedImageSizes": [320, 420, 640, 750, 768, 828, 1024, 1080, 1200, 1920, 2048, 3840],
+        "allowedImageSizes": [320, 420, 640, 750, 768, 828, 1020, 1024, 1080, 1200, 1920, 2048, 3840],
```
jamesricky, line 9: `1024` is probably not needed then... maybe even all other values above it. 

Since this is only used for this specific mail (as far as I know), maybe it makes sense to remove all values that are not common image sizes in the mail 🤔 

What do you think @raphaelblum?

demo/campaign/src/common/blocks/CommonImageBlock.tsx
```diff
@@ -0,0 +1,74 @@
...
+    <MjmlStyle>{css`
+        @media (max-width: ${theme.mailSize.contentWidth - 1}px) {
+            .image-block > table > tbody > tr > td {
+                width: 100% !important;
+            }
+
+            .image-block img {
+                height: auto !important;
+            }
+        }
```
jamesricky, line 26: The responsive behavior doesn't work correctly. 
Please make sure the aspect ratio is always used correctly and that the image is scaled down, as soon, as the width of the window is lower than the width of the mail body. 

https://github.com/user-attachments/assets/1d2e242e-ec16-44a2-8f10-80f9db43036b

demo/campaign/src/common/blocks/CommonImageBlock.tsx
```diff
@@ -0,0 +1,74 @@
...
+
+    const desktopImageHeight = Math.round(desktopRenderWidth / usedAspectRatio);
+
+    const imageClassNames = ["image-block"];
+
+    return (
+        <MjmlImage
+            src={imageUrl}
+            fluidOnMobile="true"
+            cssClass={imageClassNames.join(" ")}
```
jamesricky, line 66: I don't think having this array adds any benefit. 

```suggestion
    return (
        <MjmlImage
            src={imageUrl}
            fluidOnMobile="true"
            cssClass="image-block"
```

demo/campaign/src/common/helpers/imageBlockHelpers.ts
```diff
@@ -0,0 +1,23 @@
...
+    .map((value) => parseInt(value))
+    .sort((a, b) => a - b);
+
+export const getDamAllowedImageWidth = (minimumWidth: number): number => {
+    let width: number | null = null;
+    const largestPossibleWidth = validSizes[validSizes.length - 1];
+
+    validSizes.forEach((validWidth) => {
+        if (minimumWidth === 560) {
+            width = 1020;
```
jamesricky, line 12: Are these values that are (or should be) defined in the theme?

demo/api/src/comet-config.json
raphaelblum: We don't have any other images in the demo. I guess the allowedImageSizes were copied at project setup.

> maybe it makes sense to remove all values that are not common image sizes in the mai

Yes, I think it would make sense to remove all image sizes that are not common for mails. I don't think it makes sense having so much different imageSize options here.

demo/campaign/src/common/blocks/CommonImageBlock.tsx
juliawegmayr: The incorrect behaviour was caused by missing styling of the `IndentedSectionGroup`. 
Styling was fixed in this commit: https://github.com/vivid-planet/comet-brevo-module/pull/57/commits/712bd92a60af1863c8204bf41014e2f3b01e5be7

demo/campaign/src/components/IndentedSectionGroup.tsx
```diff
@@ -7,18 +7,31 @@ export interface IndentedSectionGroupProps extends React.ComponentProps<typeof M
     cssClass?: string;
 }
 
-const CONTENT_WIDTH = theme.mailSize.contentWidth - theme.mailSize.contentSpacing * 2;
+export const indentedSectionGroupStyles = (
+    <MjmlStyle>{`
```
jamesricky, line 11: Please use the css helper for better highlighting and prettier formatting

```suggestion
    <MjmlStyle>{css`
```

demo/campaign/src/components/IndentedSectionGroup.tsx
```diff
@@ -7,18 +7,31 @@ export interface IndentedSectionGroupProps extends React.ComponentProps<typeof M
...
+      @media (max-width: 759px) {
```
jamesricky, line 12: The default mail-width is `600px`. 
At this size and above, the default inline-styling should be all that's needed, as mail-clients do not reliably apply css. 

So the responsive behavior should only be applied below the default mail width, in this case `599px`. 

```suggestion
      @media (max-width: ${theme.mailSize.mailWidth - 1}px) {
```

demo/campaign/src/components/IndentedSectionGroup.tsx
```diff
@@ -7,18 +7,31 @@ export interface IndentedSectionGroupProps extends React.ComponentProps<typeof M
...
+        .indented-section__group {
+          width: 100% !important;
+          max-width: 592px;
```
jamesricky, line 15: Shouldn't this be the `contentWidth` from the theme?

demo/campaign/src/components/IndentedSectionGroup.tsx
```diff
@@ -7,18 +7,31 @@ export interface IndentedSectionGroupProps extends React.ComponentProps<typeof M
...
+    `}</MjmlStyle>
+);
+
+const CONTENT_WIDTH = theme.mailSize.contentWidth;
 
 export const IndentedSectionGroup = ({ children, cssClass, ...restProps }: IndentedSectionGroupProps) => {
     return (
         <MjmlSection
-            cssClass={`${cssClass || ""}`.trim()}
+            cssClass={(`${cssClass || ""}`.trim(), "indented-section")}
```
jamesricky, line 26: I think this needs to be one string 🤔 

```suggestion
            cssClass={`indented-section ${cssClass || ""}`.trim()}
```

demo/campaign/src/components/IndentedSectionGroup.tsx
```diff
@@ -7,18 +7,31 @@ export interface IndentedSectionGroupProps extends React.ComponentProps<typeof M
...
             paddingLeft={theme.mailSize.contentSpacing}
             paddingRight={theme.mailSize.contentSpacing}
             backgroundColor={theme.colors.background.content}
             {...restProps}
         >
-            <MjmlGroup width={CONTENT_WIDTH}>{children}</MjmlGroup>
+            <MjmlGroup width={CONTENT_WIDTH} cssClass="indented-section__group">
```
jamesricky, line 32: I'd just use the theme variable directly here, without defining an additional variable. 

```suggestion
            <MjmlGroup width={theme.mailSize.contentWidth} cssClass="indented-section__group">
```

demo/campaign/src/common/blocks/CommonImageBlock.tsx
jamesricky: The aspect ratio still changes, when the responsive styling is applied.

demo/campaign/src/common/helpers/imageBlockHelpers.ts
```diff
@@ -0,0 +1,23 @@
...
+    .map((value) => parseInt(value))
+    .sort((a, b) => a - b);
+
+export const getDamAllowedImageWidth = (minimumWidth: number, contentWidth): number => {
+    let width: number | null = null;
+    const largestPossibleWidth = validSizes[validSizes.length - 1];
+
+    validSizes.forEach((validWidth) => {
+        if (minimumWidth === contentWidth) {
+            width = 1020;
```
jamesricky, line 12: Shouldn't this be exactly twice the `contentWidth`? 
The `contentWidth` is exactly the maximum width the image can be rendered in, correct?

juliawegmayr:
@jamesricky As discussed last week, I removed the padding of the image, so that it takes the full width of the defined contentWidth. 

Uploading Screen Recording 2024-08-16 at 10.05.14.mov…

demo/campaign/src/components/Root.tsx
```diff
@@ -33,6 +36,8 @@ export const Root: React.FC<Props> = ({ children, title }) => {
         <Mjml>
             <MjmlHead>
                 {!!title && <MjmlTitle>{title}</MjmlTitle>}
```
jamesricky, line 38: The text inside `Typography` and also the `Divider` have a side-padding by default, in addition to the padding created by `IndentedSectionGroup`. 

image

It should probably only include the padding from `IndentedSectionGroup`, right?
You could remove the default-padding from the Mjml components by adding the following to `MjmlHead`. 

```tsx
<MjmlAttributes>
    <MjmlText padding={0} />
    <MjmlDivider padding={0} />
</MjmlAttributes>
```

demo/campaign/src/blocks/ImageBlock.tsx
```diff
@@ -0,0 +1,16 @@
+import { PropsWithData } from "@comet/cms-site";
+import { MjmlColumn } from "@luma-team/mjml-react";
+import { PixelImageBlockData } from "@src/blocks.generated";
+import { CommonImageBlock } from "@src/common/blocks/CommonImageBlock";
+import { IndentedSectionGroup } from "@src/components/IndentedSectionGroup";
+import { theme } from "@src/util/theme";
+
+export const ImageBlock = ({ data }: PropsWithData<PixelImageBlockData>) => {
+    return (
+        <IndentedSectionGroup paddingLeft={0} paddingRight={0}>
```
jamesricky, line 10: If you don't want the indentation, please use `MjmlSection` instead. 
It does have a default-padding but that should probably be removed globally, like here: ...

demo/campaign/src/common/blocks/CommonImageBlock.tsx
```diff
@@ -0,0 +1,63 @@
...
+            padding: 0 !important;
+        }
+
+        .image-block > table > tbody > tr > td {
+            width: 100% !important;
+        }
+
+        .image-block img {
+            height: auto !important;
+        }
```
jamesricky, line 27: CSS will not be loaded on all mail-clients, the styling needs to work in that case as well, at least the desktop-variant. Please only use CSS for mobile/responsive styling and use inline-styles for everything else.

demo/campaign/src/blocks/ImageBlock.tsx
jamesricky: > It does have a default-padding but that should probably be removed globally, like here: https://github.com/vivid-planet/comet-brevo-module/pull/57#discussion_r1721267201

This is still missing, right? 
There is now a spacing above and below the image:

Screenshot 2024-08-19 at 10 17 49

juliawegmayr: I think, I got it now - thank you @jamesricky  👍

jamesricky (approved):

raphaelblum (approved):

PR #59, opened 2024-07-17
COM-687 Add automatic deletion process

juliawegmayr:

All contacts that are set to “Blocked” for email campaigns should be completely removed from the contacts using a regular deletion job in order to be GDPR compliant.


packages/api/src/brevo-contact/brevo-contact.console.ts
```diff
@@ -0,0 +1,53 @@
...
+                    domain: targetGroup.scope.domain,
+                    language: targetGroup.scope.language,
+                });
+
+                const blacklistedContacts = contacts.filter((contact) => contact.emailBlacklisted === true);
+
+                if (blacklistedContacts.length > 0) {
+                    await this.brevoApiContactsService.deleteContacts(blacklistedContacts, {
+                        domain: "main",
+                        language: "en",
```
RainbowBunchie, line 44: This should be 
```
 domain: targetGroup.scope.domain,
 language: targetGroup.scope.language,
 ```

File: packages/api/src/brevo-contact/brevo-contact.console.ts [hunk 2]
```diff
@@ -0,0 +1,53 @@
...
+    @CreateRequestContext()
+    async execute(): Promise<void> {
+        const offset = 0;
+        const limit = 50;
+        const where = { isMainList: true };
+
+        const [targetGroups] = await this.targetGroupsService.findMainTargetGroups({ offset, limit, where });
+
+        for (const targetGroup of targetGroups) {
+            let numberOfBlacklistedContacts = false;
```
RainbowBunchie, line 30: As the variable naming suggests I would assume this should be a number. 
So this should always be a number and be `0` here

packages/api/src/brevo-contact/brevo-contact.console.ts
```diff
@@ -0,0 +1,53 @@
...
+                const blacklistedContacts = contacts.filter((contact) => contact.emailBlacklisted === true);
+
+                if (blacklistedContacts.length > 0) {
+                    await this.brevoApiContactsService.deleteContacts(blacklistedContacts, {
+                        domain: "main",
+                        language: "en",
+                    });
+                }
+
+                numberOfBlacklistedContacts = contacts.length === limit;
```
RainbowBunchie, line 48: This is not the number of blocklisted contacts. 
The variable should be renamed

packages/api/src/brevo-contact/brevo-contact.console.ts
```diff
@@ -0,0 +1,53 @@
...
+                offset += limit;
+            } while (numberOfBlacklistedContacts);
```
RainbowBunchie, line 50: Also I think this condition is generally wrong 
`contacts.length === limit` will always be true in the first loop run if there are more than 50 contacts, right? 
I think this should be `contacts.length < limit` because then there are no more contacts to fetch

packages/api/src/target-group/target-groups.service.ts
```diff
@@ -106,22 +106,18 @@ export class TargetGroupsService {
         return true;
     }
 
-    async findNonMainTargetGroups({
+    async findMainTargetGroups({
```
RainbowBunchie, line 109: This should be renamed to `findTargetGroups` since it is possible to find main and also non main target groups with this function

packages/api/src/target-group/target-groups.service.ts
```diff
@@ -106,22 +106,18 @@ export class TargetGroupsService {
...
         scope,
         offset,
         limit,
+        where,
     }: {
         scope?: EmailCampaignScopeInterface;
         offset: number;
         limit: number;
+        where: Partial<TargetGroupInterface>;
```
RainbowBunchie, line 118:
```
where:  FilterQuery<TargetGroupInterface>;
```

.changeset/cold-clouds-hang.md
```diff
@@ -0,0 +1,7 @@
+---
+"@comet/brevo-api": minor
+---
+
+Add `DeleteUnsubscribedContactsConsole` job to enable the deletion of blocklisted contacts. This job can be utilized as a cronjob to periodically clean up the blocklisted contacts.
+
+Add additional functions `findContacts` and `deleteContacts` in `brevo-api-contact-service-ts` to enable finding and deleting all contacts.
```
RainbowBunchie, line 7: In my opinion this information should be deleted, since these functions cannot be used outside the package since the brevoApiContactService is not exported so this information is not useful. @raphaelblum what is your opinion on this? Should this be included in the changeset?  

```suggestion
```

RainbowBunchie (changes requested):

packages/api/src/brevo-contact/brevo-contact.console.ts
```diff
@@ -0,0 +1,53 @@
...
+        for (const targetGroup of targetGroups) {
+            let hasMoreContacts = false;
+            let offset = 0;
+
+            do {
+                const contacts = await this.brevoApiContactsService.findContacts(limit, offset, {
+                    domain: targetGroup.scope.domain,
+                    language: targetGroup.scope.language,
+                });
+
```
RainbowBunchie, line 38: I just realized that you cannot set domain and language here like that. Because the scope could also be composed of different props depending on the project, the scope is not defined in the module. 
can you change it to something like
```
const contacts = await this.brevoApiContactsService.findContacts(limit, offset, {
                    scope: targetGroup.scope,
                });
                ```

File: .changeset/cold-clouds-hang.md (hunk 7 above)
Comment (raphaelblum) [reply .changeset/cold-clouds-hang.md:7]: yes, I'd say only the relevant info for the application that uses the packages should be in the changeset.

File: packages/api/src/target-group/target-groups.service.ts [hunk 9]
```diff
@@ -106,22 +106,18 @@ export class TargetGroupsService {
...
     }
 
-    async findNonMainTargetGroups({
+    async findTargetGroups({
         scope,
         offset,
         limit,
+        where,
     }: {
         scope?: EmailCampaignScopeInterface;
```
raphaelblum, line 115: The scope is not used anymore.

I would remove it here, but pass it in the "where" prop

packages/api/src/brevo-contact/brevo-contact.console.ts
```diff
@@ -0,0 +1,49 @@
...
+                    scope: targetGroup.scope,
+                });
+
+                const blacklistedContacts = contacts.filter((contact) => contact.emailBlacklisted === true);
+
+                if (blacklistedContacts.length > 0) {
+                    await this.brevoApiContactsService.deleteContacts(blacklistedContacts, { scope: targetGroup.scope });
+                }
+
+                hasMoreContacts = contacts.length > limit;
```
RainbowBunchie, line 44: The contacts.length can never be bigger than the limit you set for the contacts. It can only be smaller than the limit
This should be `!(contacts.length < limit)` I think

raphaelblum (approved):

RainbowBunchie (approved):

PR #60, opened 2024-07-23
COM-726: Reduce brevo requests

RainbowBunchie:
COM-726

Problem: 
Brevo allows only limited amount of req/h for the campaigns endpoint

Measures taken to reduce the api requests:
- Save sending state in database instead of always fetching live from brevo -> only fetching for scheduled campaigns with scheduledAt date in the past is necessary  
- Cache api response for campaigns for 1 minute before fetching again from brevo (relevant for sending state and statistics)

### DIFF
--- a/packages/api/src/brevo-api/brevo-api.module.ts
+++ b/packages/api/src/brevo-api/brevo-api.module.ts
@@ -1,3 +1,4 @@
+import { CacheModule } from "@nestjs/cache-manager";
 import { Module } from "@nestjs/common";
 
 import { ConfigModule } from "../config/config.module";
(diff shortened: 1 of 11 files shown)


packages/api/src/brevo-api/brevo-api.module.ts
```diff
@@ -1,11 +1,12 @@
+import { CacheModule } from "@nestjs/cache-manager";
 import { Module } from "@nestjs/common";
 
 import { ConfigModule } from "../config/config.module";
 import { BrevoApiCampaignsService } from "./brevo-api-campaigns.service";
 import { BrevoApiContactsService } from "./brevo-api-contact.service";
 
 @Module({
-    imports: [ConfigModule],
+    imports: [ConfigModule, CacheModule.register({ ttl: 1000 * 60 })],
```
raphaelblum, line 9: according to the docs, ttl is in seconds, see: https://docs.nestjs.com/techniques/caching#customize-caching

I think we said we want the cache to be 1 Minute, so I would define it as ttl: 60

PR #63, opened 2024-07-30
COM-965 Add grid to display all contacts in a target group

juliawegmayr:

Add datagrid to display all assigned contacts of a target group.

Screenshot 2024-07-30 at 08 17 27


demo/api/schema.gql
```diff
@@ -518,7 +518,7 @@ type Query {
   mainMenu(scope: PageTreeNodeScopeInput!): [PageTreeNode!]!
   brevoContact(id: Int!, scope: EmailCampaignContentScopeInput!): BrevoContact!
   brevoContacts(targetGroupId: ID, email: String, scope: EmailCampaignContentScopeInput!, offset: Int! = 0, limit: Int! = 25): PaginatedBrevoContacts!
-  manuallyAssignedBrevoContacts(offset: Int! = 0, limit: Int! = 25, targetGroupId: ID!, email: String): PaginatedBrevoContacts!
+  assignedBrevoContacts(offset: Int! = 0, limit: Int! = 25, targetGroupId: ID!, email: String, brevoId: Float): PaginatedBrevoContacts!
```
RainbowBunchie, line 521: brevo id should be an integer (not float), please check why the type is wrong here

packages/api/src/brevo-contact/brevo-contact.resolver.ts
```diff
@@ -93,10 +93,11 @@ export function createBrevoContactResolver({
         }
 
         @Query(() => PaginatedBrevoContacts)
-        async manuallyAssignedBrevoContacts(
-            @Args() { offset, limit, email, targetGroupId }: ManuallyAssignedBrevoContactsArgs,
+        async assignedBrevoContacts(
+            @Args() { offset, limit, email, targetGroupId, brevoId }: AssignedBrevoContactsArgs,
         ): Promise<PaginatedBrevoContacts> {
             const targetGroup = await this.targetGroupRepository.findOneOrFail({ id: targetGroupId });
+            const brevoListId = brevoId ? targetGroup.brevoId : targetGroup.assignedContactsTargetGroupBrevoId;
```
RainbowBunchie, line 100: I do not think it is a good idea to send the brevoId and the targetGroup id, especially since you are not using the brevo id here eitherway. I would suggest to add a boolean maybe 'onlyManuallyAssigned' and use this variable here.

packages/admin/src/targetGroups/allAssignedContacts/AssignedContactsGrid.tsx
```diff
@@ -0,0 +1,124 @@
+/* eslint-disable @comet/no-private-sibling-import */
```
RainbowBunchie, line 1: This rule should not be disabled. Why is this neccessary here?

packages/admin/src/targetGroups/allAssignedContacts/AssignedContactsGrid.tsx
```diff
@@ -0,0 +1,124 @@
...
+    );
+};
+
+interface AllAssignedContactsGridProps {
+    scope: ContentScopeInterface;
+    id: string;
+    brevoId?: number;
+}
+
+export function AssignedContactsGrid({ id, scope, brevoId }: AllAssignedContactsGridProps): React.ReactElement {
```
RainbowBunchie, line 48:
```
AllAssignedContactsGrid
```

packages/admin/src/targetGroups/addContacts/AddContactsGridSelect.gql.ts
```diff
@@ -11,9 +11,9 @@ const targetGroupBrevoContactsFragment = gql`
     }
 `;
 
-export const manuallyAssignedBrevoContactsGridQuery = gql`
-    query ManuallyAssignedBrevoContactsGrid($offset: Int, $limit: Int, $email: String, $targetGroupId: ID!) {
-        manuallyAssignedBrevoContacts(offset: $offset, limit: $limit, email: $email, targetGroupId: $targetGroupId) {
+export const assignedBrevoContactsGridQuery = gql`
+    query AssignedBrevoContactsGrid($offset: Int, $limit: Int, $email: String, $targetGroupId: ID!) {
```
RainbowBunchie, line 15: I would keep the `ManuallyAssignedBrevoContactsGrid` naming to easier differentiate between the 2 similar queries (all & manually)

packages/admin/src/targetGroups/addContacts/AddContactsGridSelect.tsx
```diff
@@ -95,7 +95,7 @@ const useSubmitMutation = (id: string) => {
     const [addContactsToTargetGroup] = useMutation<GQLAddBrevoContactsToTargetGroupMutation, GQLAddBrevoContactsToTargetGroupMutationVariables>(
         addBrevoContactsToTargetGroupMutation,
         {
-            refetchQueries: [namedOperations.Query.ManuallyAssignedBrevoContactsGrid, targetGroupFormNamedOperations.Query.TargetGroupForm],
+            refetchQueries: [namedOperations.Query.AssignedBrevoContactsGrid, targetGroupFormNamedOperations.Query.TargetGroupForm],
```
RainbowBunchie, line 98: Does the allAssignedBrevoContactsQuery also need to be refetched here? So the grid updates when the assignment changes?

packages/admin/src/targetGroups/allAssignedContacts/AssignedContactsGrid.gql.ts
```diff
@@ -0,0 +1,24 @@
...
+        createdAt
+        modifiedAt
+        email
+        emailBlacklisted
+        smsBlacklisted
+    }
+`;
+
+export const assignedBrevoContactsGridQuery = gql`
+    query AssignedBrevoContactsGrid($offset: Int, $limit: Int, $email: String, $targetGroupId: ID!, $brevoId: Float) {
```
RainbowBunchie, line 15: I would suggest to name this Query `AssignedBrevoContactsGrid` to easier differentiate between the 2 similar queries

RainbowBunchie (changes requested):

packages/api/src/brevo-contact/dto/assigned-brevo-contacts.args.ts
```diff
@@ -1,14 +1,19 @@
...
+export class AssignedBrevoContactsArgs extends OffsetBasedPaginationArgs {
     @Field(() => ID)
     @IsString()
     targetGroupId: string;
 
     @Field(() => String, { nullable: true })
     @IsOptional()
     email?: string;
+
+    @Field(() => Number, { nullable: true })
```
raphaelblum, line 15: @Field(() => Int)

Then the type should be Integer and not float

packages/admin/src/targetGroups/TargetGroupForm.tsx
```diff
@@ -175,19 +176,36 @@ export function TargetGroupForm({ id, scope, additionalFormFields, input2State,
...
+                                </FieldSet>
+                                <FieldSet
+                                    title={
+                                        <FormattedMessage
+                                            id="cometBrevoModule.targetGroup.allAssignedContacts"
+                                            defaultMessage="All assigned contacts"
+                                        />
+                                    }
+                                    disablePadding
+                                    initiallyExpanded={false}
```
raphaelblum, line 204: not important but this is probably unnecessary, i guess initiallyExpanded={false} is the default, and you just set initiallyExpanded when you need it expanded

packages/admin/src/targetGroups/allAssignedContacts/AssignedContactsGrid.tsx
juliawegmayr: Thank you! 
I used the wrong import here.

packages/admin/src/targetGroups/addContacts/AddContactsGridSelect.tsx
juliawegmayr: Good point! 
But I tested it, and the grid already updates, although I can't explain why...

packages/admin/src/targetGroups/allAssignedContacts/AssignedContactsGrid.gql.ts
juliawegmayr: I changed it to `AllAssignedBrevoContactsGrid` (I guess, you meant that as well :) )

demo/api/schema.gql
juliawegmayr: Thanks :) 
It's because of the wrong type in the args, as Raphi pointed out.

packages/admin/src/targetGroups/allAssignedContacts/AssignedContactsGrid.gql.ts
RainbowBunchie: Yes that's exactly what i ment! I added the 'All' in my mind I guess 😂

packages/admin/src/targetGroups/addContacts/AddContactsGridSelect.tsx
RainbowBunchie: If it works, it works 🤷🏻‍♀️  👍🏻

packages/admin/src/targetGroups/addContacts/AddContactsGridSelect.gql.ts
```diff
@@ -12,8 +12,14 @@ const targetGroupBrevoContactsFragment = gql`
...
 export const manuallyAssignedBrevoContactsGridQuery = gql`
-    query ManuallyAssignedBrevoContactsGrid($offset: Int, $limit: Int, $email: String, $targetGroupId: ID!) {
-        manuallyAssignedBrevoContacts(offset: $offset, limit: $limit, email: $email, targetGroupId: $targetGroupId) {
+    query ManuallyAssignedBrevoContactsGrid($offset: Int, $limit: Int, $email: String, $targetGroupId: ID!, $onlyManuallyAssigned: Boolean) {
+        assignedBrevoContacts(
+            offset: $offset
+            limit: $limit
+            email: $email
+            targetGroupId: $targetGroupId
+            onlyManuallyAssigned: $onlyManuallyAssigned
```
RainbowBunchie, line 21: You could set this to `true` here directly since this query should only be used for manuasllyAssignedContacts as the naming suggests

(same for the all query)

packages/api/src/brevo-contact/brevo-contact.resolver.ts
```diff
@@ -93,10 +93,11 @@ export function createBrevoContactResolver({
         }
 
         @Query(() => PaginatedBrevoContacts)
-        async manuallyAssignedBrevoContacts(
-            @Args() { offset, limit, email, targetGroupId }: ManuallyAssignedBrevoContactsArgs,
+        async assignedBrevoContacts(
+            @Args() { offset, limit, email, targetGroupId, onlyManuallyAssigned }: AssignedBrevoContactsArgs,
```
RainbowBunchie, line 97: I just realised that there were already 2 different queries before for manually assigned and all brevo contacts in a target group. 

the brevo contacts query has a targetGroupId argument. So you can get all assigned contacts with this query 
 ```
 brevoContacts(targetGroupId: $targetGroupId, offset: $offset, limit: $limit, scope: $scope) {  
 ```
 and the `manuallyAssignedBrevoContacts` should just be used for the manually assigned contacts. 
 
 Should have remembered this sooner, sorry!

RainbowBunchie (approved):

raphaelblum (approved):

PR #64, opened 2024-07-30
COM-962: Improve UI/UX of target group page 

juliawegmayr:

- Add dialog to create a new target group. After creating, the user will be redirected to the editPage. 
- Setting filters is now optional, when creating a target group. 

The possibility to pass no filters at all, still has to be checked and will be added in a follow-up PR. 

https://github.com/user-attachments/assets/9d81016a-a535-46dd-9299-0b6d2884d918


packages/admin/src/targetGroups/TargetGroupDialog.tsx
```diff
@@ -0,0 +1,98 @@
...
+import { GQLCreateTargetGroupMutation, GQLCreateTargetGroupMutationVariables } from "./TargetGroupDialog.gql.generated";
+import { EditTargetGroupFinalFormValues } from "./TargetGroupForm";
+
+interface FormProps {
+    open: boolean;
+    handleClose: () => void;
+    scope: ContentScopeInterface;
+}
+
+export function TargetGroupDialog({ scope, open, handleClose }: FormProps): React.ReactElement {
```
RainbowBunchie, line 20:
```
export function TargetGroupDialog({ scope, open, handleClose }: TargetGroupDialogProps): React.ReactElement {
```

packages/admin/src/targetGroups/TargetGroupDialog.tsx
```diff
@@ -0,0 +1,98 @@
...
+            const id = mutationResponse?.createTargetGroup.id;
+            if (id) {
+                setTimeout(() => {
+                    stackSwitchApi.activatePage("edit", id);
+                });
+            }
+        }
+    };
+    return (
+        <Dialog open={open} onClose={handleClose} maxWidth="sm">
```
RainbowBunchie, line 49: Comet has an EditDialog component which could be used here 
https://comet-admin.netlify.app/?path=/story/stories-components-editdialog-edit-dialog--edit-dialog

RainbowBunchie:
@juliawegmayr you should also remove the create from the previous create/update form since the create should only be possible with the dialog right?

RainbowBunchie (changes requested):

raphaelblum (dismissed):

packages/admin/src/targetGroups/TargetGroupForm.tsx
```diff
@@ -103,28 +99,13 @@ export function TargetGroupForm({ id, scope, additionalFormFields, input2State,
...
-            });
-            if (!event.navigatingBack) {
-                const id = mutationResponse?.createTargetGroup.id;
-                if (id) {
-                    setTimeout(() => {
-                        stackSwitchApi.activatePage("edit", id);
-                    });
-                }
-            }
+        if (!id) {
```
RainbowBunchie, line 102: The id cannot be undefined here right? Should be required in the props.

RainbowBunchie (dismissed):

raphaelblum (approved):

raphaelblum (approved):

PR #65, opened 2024-07-30
COM-990: Allow configuration of redirect url

juliawegmayr:

Allow configuration of redirect urls depending on their scopes


packages/api/src/brevo-contact/validator/redirect-url.validator.ts
```diff
@@ -22,14 +23,22 @@ export const IsValidRedirectURL = (validationOptions?: ValidationOptions) => {
...
+        const [scope] = args.constraints;
+
+        try {
+            const configForScope = this.config.brevo.resolveConfig(scope);
+            if (configForScope && urlToValidate?.startsWith(configForScope.allowedRedirectUrl)) {
+                return true;
+            }
+        } catch (error) {
+            console.error(`Error resolving config for scope: ${scope}`, error);
         }
```
raphaelblum, line 36: Why is there a try-catch block here? Which part could throw an error?

In the case of an unsupported scope, wouldn't it be better to throw an error?

raphaelblum (approved):

PR #67, opened 2024-08-08
Add transactional mail service

raphaelblum:

COM-966

added a controller in the demo as example and that works with this request:
Screenshot 2024-08-08 at 13 57 41

### DIFF
--- a/.changeset/sour-yaks-jog.md
+++ b/.changeset/sour-yaks-jog.md
@@ -10,7 +10,6 @@ Add and export `BrevoTransactionalMailsService` that can be used in the applicat
 constructor(private readonly brevoTransactionalMailsService: BrevoTransactionalMailsService) {}
 
 async send(email: string, htmlContent: string, subject: string): Promise<void> {
-    const { text } = data;
     await this.brevoTransactionalMailsService.send({ to: [{ email }], htmlContent, subject }, data.scope);
 }
 ```
--- a/.changeset/sour-yaks-jog.md
+++ b/.changeset/sour-yaks-jog.md
@@ -1,0 +1,16 @@
+---
+"@comet/brevo-api": minor
+---
+
+Add and export `BrevoTransactionalMailsService` that can be used in the application for sending transactional mails.
+
+**Example Usage of `BrevoTransactionalMailsService`**
+
+```typescript
+constructor(private readonly brevoTransactionalMailsService: BrevoTransactionalMailsService) {}
+
+async send(email: string, htmlContent: string, subject: string): Promise<void> {
+    const { text } = data;
+    await this.brevoTransactionalMailsService.send({ to: [{ email }], htmlContent, subject }, data.scope);
+}
+```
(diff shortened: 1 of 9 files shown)


.changeset/sour-yaks-jog.md
```diff
@@ -0,0 +1,15 @@
...
+
+Add and export `BrevoTransactionalMailsService` that can be used in the application for sending transactional mails.
+
+**Example Usage of `BrevoTransactionalMailsService`**
+
+```typescript
+constructor(private readonly brevoTransactionalMailsService: BrevoTransactionalMailsService) {}
+
+async send(email: string, htmlContent: string, subject: string): Promise<void> {
+    await this.brevoTransactionalMailsService.send({ to: [{ email }], htmlContent, subject }, data.scope);
```
RainbowBunchie, line 13: In the code htmlContent is calles textContent, which one is correct? Does is have to be html?

demo/api/src/brevo-transactional-mails/brevo-transactional-mails.controller.ts
```diff
@@ -0,0 +1,17 @@
...
+@Controller("transactional-mails")
+export class BrevoTransactionalMailsController {
+    constructor(private readonly brevoTransactionalMailsService: BrevoTransactionalMailsService) {}
+
+    @DisableGlobalGuard()
+    @Post(`/send`)
+    async send(@Body() data: BrevoTransactionalMailsBody): Promise<void> {
+        const { text } = data;
+        await this.brevoTransactionalMailsService.send({ to: [{ email: data.to }], textContent: text, subject: data.subject }, data.scope);
+    }
```
RainbowBunchie:
```
        const { text, subject, to, scope } = data;
        await this.brevoTransactionalMailsService.send({ to: [{ email: to }], textContent: text, subject:subject },scope);
    }
```

packages/api/src/brevo-api/brevo-api-transactional-mails.service.ts
```diff
@@ -0,0 +1,36 @@
...
+
+        this.transactionalEmailsApi.set(JSON.stringify(scope), transactionalEmailApi);
+
+        return transactionalEmailApi;
+    }
+
+    async send(options: Omit<Brevo.SendSmtpEmail, "sender">, scope: EmailCampaignScopeInterface): SendTransacEmailResponse {
+        const config = this.config.brevo.resolveConfig(scope);
+        return this.getTransactionalEmailsApi(scope).sendTransacEmail({ ...options, sender: config.sender });
+    }
```
RainbowBunchie, line 35: I do not know what props are in `Brevo.SendSmtpEmail`, but do we want all of them to be adaptable?

.changeset/sour-yaks-jog.md
RainbowBunchie: This is connected to the other concern, see below

raphaelblum: Emails can be sent with html or text or both.

Text often was used as the preview in email clients. Afaik all new email clients now also support htmlContent as preview. 
But in general you can set both, but I guess most projects will use htmlContent, unless its just some reporting where text is sufficient.

packages/api/src/brevo-api/brevo-api-transactional-mails.service.ts
RainbowBunchie: Yes then it is fine like that 👍🏻

RainbowBunchie (approved):

PR #71, opened 2024-08-13
Import contacts from CSV

thomashuettmaier:

Implement CSV Brevo Contact import in Admin
Implement console command to upload CSV Brevo Contacts

TIMR: COM-1001 - Multi Select für Target Group Auswahl beim Versenden
JIRA: https://vivid-planet.atlassian.net/browse/COM-1001

TIMR: COM-967 - CSV Import für Kontakte - via Console Command
JIRA: https://vivid-planet.atlassian.net/browse/COM-967

---

Importing to contacts:

https://github.com/user-attachments/assets/216178f0-cdd3-4f81-a3b6-7ccc1e52c1bc

Importing to target group:

https://github.com/user-attachments/assets/5b57a04b-4d73-4cd5-a08f-15603a7643c9


packages/admin/src/common/form/CsvUploadField.tsx
```diff
@@ -0,0 +1,81 @@
...
+    submitting: boolean;
+}
+
+export const CsvUploadField: React.FC<CsvUploadFieldProps> = ({ buttonText, submitting, input: { onChange, value } }) => {
+    return (
+        <>
+            {submitting && (
+                <LoadingWrapper>
+                    <CircularProgress />
+                </LoadingWrapper>
```
johnnyomair, line 20: Please use the Loading component from `@comet/admin`

packages/admin/src/common/upload.ts
```diff
@@ -0,0 +1,17 @@
+import { createConfig } from "../config";
+import { GQLEmailCampaignContentScopeInput } from "../graphql.generated";
+
+const config = createConfig();
+
+export function upload(file: File, scope: GQLEmailCampaignContentScopeInput, listIds?: string[]): Promise<Response> {
```
johnnyomair, line 6: This shouldn't be a common file if it uploads to a specific endpoint.

packages/admin/src/common/upload.ts
```diff
@@ -0,0 +1,17 @@
+import { createConfig } from "../config";
+import { GQLEmailCampaignContentScopeInput } from "../graphql.generated";
+
+const config = createConfig();
```
johnnyomair, line 4: There shouldn't be a createConfig function in the library. This should come via injection (= a React provider) from the application.

.changeset/sharp-trees-yawn.md
```diff
@@ -0,0 +1,18 @@
+---
+"@comet/brevo-api": major
+"@comet/brevo-admin": minor
+---
+
+Added functionality to import Brevo contacts from CSV files.
+
+**Important Notice:**
+
+-   A major bump in `@comet/brevo-api` is required because a new `.env` variable named `REDIRECT_URL_FOR_IMPORT` is needed.
```
johnnyomair, line 10: Why is this redirect URL necessary? Why can't this be specified as CLI arg during the import?

packages/api/package.json
```diff
@@ -25,10 +25,14 @@
         "test:watch": "jest --watch"
     },
     "dependencies": {
+        "@fast-csv/parse": "^5.0.0",
         "@getbrevo/brevo": "^2.2.0",
         "@nestjs/axios": "^1.0.0",
         "@nestjs/cache-manager": "^2.2.2",
+        "@nestjs/platform-express": "^9.0.0",
```
johnnyomair, line 32: This probably needs to be a peer dependency

packages/api/src/brevo-contact/brevo-contact-import.controller.ts
```diff
@@ -0,0 +1,53 @@
...
+            const contentContent = file.buffer.toString("utf8");
+            const parsedScope = JSON.parse(scope) as EmailCampaignScopeInterface;
+
+            let parsedListIds = undefined;
+            if (listIds) parsedListIds = JSON.parse(listIds) as string[];
+
+            const targetGroupBrevoIds = (await this.targetGroupRepository.find({ id: { $in: parsedListIds } })).map(({ brevoId }) => brevoId);
+
+            const redirectUrl = this.config.brevo.resolveConfig(parsedScope).redirectUrlForImport;
+            await this.brevoContactImportService.importContactFromCsv(contentContent, parsedScope, redirectUrl, targetGroupBrevoIds);
```
johnnyomair, line 48: This could in theory take longer than 30s (default request timeout). @raphaelblum  Is this something we should take into consideration?

packages/api/src/brevo-contact/brevo-contact.module.ts
```diff
@@ -34,17 +37,23 @@ export class BrevoContactModule {
...
                 BrevoContactsService,
                 BrevoContactResolver,
                 EcgRtrListService,
                 IsValidRedirectURLConstraint,
                 DeleteUnsubscribedBrevoContactsConsole,
+                BrevoContactImportConsole,
             ],
-            exports: [BrevoContactsService],
+            controllers: [BrevoContactImportController],
+            exports: [BrevoContactsService, BrevoContactImportService, BrevoContactImportService],
```
johnnyomair, line 56:
```suggestion
            exports: [BrevoContactsService, BrevoContactImportService],
```

packages/api/src/brevo-contact/contact-importer-error.interceptor.ts
```diff
@@ -0,0 +1,21 @@
+import { BadRequestException, CallHandler, ExecutionContext, Injectable, NestInterceptor } from "@nestjs/common";
+import { catchError, Observable } from "rxjs";
+
+import { ImporterError } from "../common/importer.error";
+
+@Injectable()
+export class ImporterErrorInterceptor implements NestInterceptor {
```
johnnyomair, line 7: Do we need an extra interceptor and error? Why not throw a CometValidationException?

johnnyomair (commented):
We should definitely wait for @raphaelblum's review here.

packages/api/src/brevo-contact/brevo-contact-import.controller.ts
raphaelblum: Yes, I believe that will happen to us at some point so we should take it into consideration.

Probably not for the current use case, but when I look at the function, we send 1 request for the contact and then 1 for updating or creating a double opt in contact. If I calculate with a very fast api and a request only takes 25ms, that would be 30000 ms / 50 ms  = 600 Contacts. 

And we already had imports of 1000-5000 contacts.

.changeset/sharp-trees-yawn.md
raphaelblum: Yes I agree with what @thomasdax98.

@thomashuettmaier  and I already talked about it as well, and we also thought that it probably makes the most sense to define it this way. Even though I don't like defining another variable for redirectUrl, all the other solutions we came up with were worse.

packages/api/src/brevo-contact/brevo-contact-import.service.ts
```diff
@@ -0,0 +1,169 @@
...
+@Injectable()
+export class BrevoContactImportService {
+    constructor(
+        @Inject(BREVO_MODULE_CONFIG) private readonly config: BrevoModuleConfig,
+        private readonly brevoApiContactsService: BrevoApiContactsService,
+        private readonly brevoContactsService: BrevoContactsService,
+        private readonly targetGroupsService: TargetGroupsService,
+    ) {}
+
+    async importContactFromCsv<T extends ValidateableRow>(
```
raphaelblum, line 32:
```suggestion
    async importContactsFromCsv<T extends ValidateableRow>(
```

for me, this makes more sense in plural because several contacts are imported from one file

packages/api/src/index.ts
```diff
@@ -1,4 +1,5 @@
 export { BrevoTransactionalMailsService } from "./brevo-api/brevo-api-transactional-mails.service";
+export { BrevoContactImportService } from "./brevo-contact/brevo-contact-import.service";
```
raphaelblum, line 2: I'm not sure if it makes sense exporting the importService, when it is intended to use either the .controller or the .console

What might be the use case for exporting the Import Service?

I can only think of a custom importer console or controller, but I am not sure why this might be necessary.

packages/api/src/brevo-contact/brevo-contact-import.service.ts
thomasdax98: 7b1fa97de4d56fa04f48192c9b3c3712a68fb6e6

packages/api/src/index.ts
thomasdax98: 2a661f004d4f63139d49fc34ddc5d67da59f588b

Then we'll not export it for now

packages/api/src/brevo-contact/brevo-contact-import.controller.ts
thomasdax98: da6b4d313aaa2393481a41e0112c89ed34e8f728

packages/admin/src/common/contactImport/useContactImportFromCsv.tsx
```diff
@@ -91,14 +91,24 @@ const ContactImportComponent = ({ scope, targetGroupId, fileInputRef, refetchQue
             } catch (e) {
                 setImportingCsv(false);
 
-                errorDialog?.showError({
-                    title: <FormattedMessage {...messages.error} />,
-                    userMessage: (
+                let userMessage = (
+                    <FormattedMessage
+                        id="cometBrevoModule.useContactImport.error.defaultMessage"
+                        defaultMessage="A error occured during the import. Please try again in a while or contact your administrator if the error persists."
```
raphaelblum, line 97: "An error occured"

packages/admin/src/common/BrevoConfigProvider.tsx
```diff
@@ -0,0 +1,23 @@
...
+    return <BrevoConfigContext.Provider value={value}>{children}</BrevoConfigContext.Provider>;
+};
+
+export const useBrevoConfig = (): BrevoConfig => {
+    const context = React.useContext(BrevoConfigContext);
+    if (context === undefined) {
+        throw new Error("useBrevoConfig must be used within a BrevoConfigProvider");
+    }
+    return context;
+};
```
raphaelblum, line 23: as we already talked, I like this config and I think other variables like the scopeParts should also be defined there.
Or maybe even defining everything in the BrevoConfigProvider would make it easier in the application to use, instead of here: https://github.com/vivid-planet/comet-brevo-module/blob/main/demo/admin/src/Routes.tsx#L51

packages/admin/src/common/contactImport/useContactImportFromCsv.tsx
thomasdax98:
```suggestion
                        defaultMessage="An error occured during the import. Please try again in a while or contact your administrator if the error persists."
```

packages/api/src/brevo-contact/brevo-contact-import.service.ts
```diff
@@ -0,0 +1,169 @@
...
+        private readonly brevoContactsService: BrevoContactsService,
+        private readonly targetGroupsService: TargetGroupsService,
+    ) {}
+
+    async importContactsFromCsv<T extends ValidateableRow>(
+        csvContent: string,
+        scope: EmailCampaignScopeInterface,
+        redirectUrl: string,
+        targetGroups: TargetGroupInterface[] = [],
+        validateable?: T,
```
raphaelblum, line 37: is this ever defined ?

I can't find the function with 5 parameters, so it should be always undefined, or am I missing something?

thomasdax98: This was added by @thomashuettmaier. I guess it was meant for some kind of more advanced validation logic. But apparently it was never used. I don't think we need it for now, so I removed it. We can maybe improve validation in the future, if needed

raphaelblum: yes I think we should probably improve the validation in the future, especially for custom user uploads in the csv.

raphaelblum (approved):

PR #77, opened 2024-08-23
Fix: adapt TargetGroupForm to allow removed filters

juliawegmayr:

### DIFF
--- a/packages/admin/src/targetGroups/TargetGroupForm.tsx
+++ b/packages/admin/src/targetGroups/TargetGroupForm.tsx
@@ -54,12 +54,16 @@ export function TargetGroupForm({ id, scope, additionalFormFields, input2State, 
     const mode = "edit";
     const formApiRef = useFormApiRef<EditTargetGroupFinalFormValues>();
 
-    const targetGroupFormFragment = gql`
+    let targetGroupFormFragment: DocumentNode | undefined;
+
+    if (additionalFormFields) {
+        targetGroupFormFragment = gql`
         fragment TargetGroupForm on TargetGroup {
-            ${nodeFragment ? "...".concat(nodeFragment?.name) : ""}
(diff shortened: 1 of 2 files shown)


packages/admin/src/targetGroups/TargetGroupForm.tsx
```diff
@@ -54,12 +54,16 @@ export function TargetGroupForm({ id, scope, additionalFormFields, input2State,
...
+
+    if (additionalFormFields) {
+        targetGroupFormFragment = gql`
         fragment TargetGroupForm on TargetGroup {
-            ${nodeFragment ? "...".concat(nodeFragment?.name) : ""}
+            ${nodeFragment ? "...".concat(nodeFragment.name) : ""}
         }
         ${nodeFragment?.fragment ?? ""}
     `;
+    }
```
raphaelblum, line 66: Can we add the check for nodeFragment in the if? Wouldn't look that verbose
Or should the check in general only be for nodeFragment?

```
    if (additionalFormFields && nodeFragment) {
        targetGroupFormFragment = gql`
        fragment TargetGroupForm on TargetGroup {
            ${"...".concat(nodeFragment.name)}
        }
        ${nodeFragment.fragment}
    `;
    }
```

raphaelblum: is the use of conact `${"...".concat(nodeFragment.name)}` still needed ?
Or does it also work like that?
`
...${nodeFragment.name}`

juliawegmayr: I tried it, but I got a syntax error:
`Syntax Error: Expected "{", found "}".`

raphaelblum: okay thanks for trying 👍

raphaelblum (approved):

RainbowBunchie (approved):

PR #78, opened 2024-08-29
COM-908 Add test contact list

juliawegmayr:

- Create a test contact list
- Fix: if a contact is added to the main list, it was added to all lists without filters. Added a check for manuallyAssignedContacts.
- Contacts added to a test list, don't need a double opt-in.
- If a contact is added to the test list, it is possible to add it to the main list as well. Then a DOI is sent.
- Test list is not displayed in the target group page.

### DIFF
--- a/.changeset/chilly-dogs-draw.md
+++ b/.changeset/chilly-dogs-draw.md
@@ -1,0 +1,6 @@
+---
+"@comet/brevo-admin": major
+"@comet/brevo-api": major
+---
+
+Add test contact page
--- a/.changeset/chilly-dogs-draw.md
+++ b/.changeset/chilly-dogs-draw.md
@@ -1,6 +1,6 @@
 ---
-"@comet/brevo-admin": major
-"@comet/brevo-api": major
+"@comet/brevo-admin": minor
+"@comet/brevo-api": minor
 ---
 
 Adds `createBrevoTestContactsPage` for creating a test contacts page, that is indepent from the main list.
--- a/.changeset/chilly-dogs-draw.md
+++ b/.changeset/chilly-dogs-draw.md
@@ -3,4 +3,6 @@
 "@comet/brevo-api": major
 ---
 
-Add test contact page
+Adds `createBrevoTestContactsPage` for creating a test contacts page, that is indepent from the main list.
+
+Remove `email` and `redirectionUrl` from `brevoContactsPageAttributesConfig`
--- a/demo/admin/src/common/brevoModuleConfig/brevoContactsPageAttributesConfig.tsx
+++ b/demo/admin/src/common/brevoModuleConfig/brevoContactsPageAttributesConfig.tsx
@@ -70,7 +70,7 @@ export const getBrevoContactConfig = (
         name: string;
     };
     input2State: (values?: AdditionalFormConfigInputProps) => {
-        attributes: { SALUTATION?: GQLBrevoContactSalutation; FIRSTNAME?: string; LASTNAME?: string };
+        attributes: { BRANCH?: Array<GQLBrevoContactBranch>; SALUTATION?: GQLBrevoContactSalutation; FIRSTNAME?: string; LASTNAME?: string };
     };
     exportFields: {
         renderValue: (row: GQLBrevoContactAttributesFragmentFragment) => string;
--- a/demo/admin/src/common/brevoModuleConfig/brevoContactsPageAttributesConfig.tsx
+++ b/demo/admin/src/common/brevoModuleConfig/brevoContactsPageAttributesConfig.tsx
@@ -138,10 +138,9 @@ export const getBrevoContactConfig = (
                 />
             </>
         ),
-        input2State: (values?: AdditionalFormConfigInputProps) => {
+        input2State: (values: AdditionalFormConfigInputProps) => {
             return {
-                email: values?.email ?? "",
-                redirectionUrl: values?.redirectionUrl ?? "",
+                ...values,
                 attributes: {
                     BRANCH: values?.attributes?.BRANCH ?? [],
                     SALUTATION: values?.attributes?.SALUTATION,
--- a/demo/admin/src/common/brevoModuleConfig/brevoContactsPageAttributesConfig.tsx
+++ b/demo/admin/src/common/brevoModuleConfig/brevoContactsPageAttributesConfig.tsx
@@ -70,9 +70,7 @@ export const getBrevoContactConfig = (
         name: string;
     };
     input2State: (values?: AdditionalFormConfigInputProps) => {
-        email: string;
-        redirectionUrl: string;
-        attributes: { BRANCH?: Array<GQLBrevoContactBranch>; SALUTATION?: GQLBrevoContactSalutation; FIRSTNAME?: string; LASTNAME?: string };
+        attributes: { SALUTATION?: GQLBrevoContactSalutation; FIRSTNAME?: string; LASTNAME?: string };
     };
     exportFields: {
         renderValue: (row: GQLBrevoContactAttributesFragmentFragment) => string;
@@ -140,7 +138,6 @@ export const getBrevoContactConfig = (
         ),
         input2State: (values: AdditionalFormConfigInputProps) => {
             return {
-                ...values,
                 attributes: {
                     BRANCH: values?.attributes?.BRANCH ?? [],
                     SALUTATION: values?.attributes?.SALUTATION,
--- a/demo/admin/src/common/brevoModuleConfig/brevoContactsPageAttributesConfig.tsx
+++ b/demo/admin/src/common/brevoModuleConfig/brevoContactsPageAttributesConfig.tsx
@@ -136,7 +136,7 @@ export const getBrevoContactConfig = (
                 />
             </>
         ),
-        input2State: (values: AdditionalFormConfigInputProps) => {
+        input2State: (values?: AdditionalFormConfigInputProps) => {
             return {
                 attributes: {
                     BRANCH: values?.attributes?.BRANCH ?? [],
--- a/packages/admin/src/brevoContacts/form/BrevoContactForm.tsx
+++ b/packages/admin/src/brevoContacts/form/BrevoContactForm.tsx
@@ -45,7 +45,7 @@ export type EditBrevoContactFormValues = {
     [key: string]: unknown;
 };
 
-type BaseBrevoContactFormValues = EditBrevoContactFormValues & {
+type EditBrevoContactFormValuesWithAttributes = EditBrevoContactFormValues & {
     email: string;
     redirectionUrl: string;
 };
@@ -62,7 +62,7 @@ export function BrevoContactForm({ id, scope, input2State, additionalFormFields,
     const stackApi = useStackApi();
     const client = useApolloClient();
     const mode = id ? "edit" : "add";
-    const formApiRef = useFormApiRef<BaseBrevoContactFormValues>();
+    const formApiRef = useFormApiRef<EditBrevoContactFormValuesWithAttributes>();
 
     const brevoContactFormFragment = gql`
         fragment BrevoContactForm on BrevoContact {
@@ -76,7 +76,7 @@ export function BrevoContactForm({ id, scope, input2State, additionalFormFields,
         id ? { variables: { id, scope } } : { skip: true },
     );
 
-    const initialValues = React.useMemo<Partial<BaseBrevoContactFormValues>>(() => {
+    const initialValues = React.useMemo<Partial<EditBrevoContactFormValuesWithAttributes>>(() => {
         let baseInitialValues = {
             email: "",
             redirectionUrl: "",
@@ -114,7 +114,11 @@ export function BrevoContactForm({ id, scope, input2State, additionalFormFields,
         },
     });
 
-    const handleSubmit = async (state: BaseBrevoContactFormValues, form: FormApi<BaseBrevoContactFormValues>, event: FinalFormSubmitEvent) => {
+    const handleSubmit = async (
+        state: EditBrevoContactFormValuesWithAttributes,
+        form: FormApi<EditBrevoContactFormValuesWithAttributes>,
+        event: FinalFormSubmitEvent,
+    ) => {
         if (await saveConflict.checkForConflicts()) {
             throw new Error("Conflicts detected");
         }
@@ -161,7 +165,7 @@ export function BrevoContactForm({ id, scope, input2State, additionalFormFields,
     }
 
     return (
-        <FinalForm<BaseBrevoContactFormValues> apiRef={formApiRef} onSubmit={handleSubmit} mode={mode} initialValues={initialValues}>
+        <FinalForm<EditBrevoContactFormValuesWithAttributes> apiRef={formApiRef} onSubmit={handleSubmit} mode={mode} initialValues={initialValues}>
             {({ values }) => (
                 <EditPageLayout>
                     {saveConflict.dialogs}
--- a/packages/admin/src/brevoContacts/form/BrevoContactForm.tsx
+++ b/packages/admin/src/brevoContacts/form/BrevoContactForm.tsx
@@ -42,9 +42,13 @@ import {
 } from "./BrevoContactForm.gql.generated";
 
 export type EditBrevoContactFormValues = {
+    [key: string]: unknown;
+};
+
+type BaseBrevoContactFormValues = EditBrevoContactFormValues & {
+    //Umbenennen
     email: string;
     redirectionUrl: string;
-    [key: string]: unknown;
 };
 
 interface FormProps {
@@ -59,14 +63,11 @@ export function BrevoContactForm({ id, scope, input2State, additionalFormFields,
     const stackApi = useStackApi();
     const client = useApolloClient();
     const mode = id ? "edit" : "add";
-    const formApiRef = useFormApiRef<EditBrevoContactFormValues>();
+    const formApiRef = useFormApiRef<BaseBrevoContactFormValues>();
 
     const brevoContactFormFragment = gql`
         fragment BrevoContactForm on BrevoContact {
             email
-            createdAt
-            emailBlacklisted
-            smsBlacklisted
             ${additionalAttributesFragment ? "...".concat(additionalAttributesFragment?.name) : ""}
         }
         ${additionalAttributesFragment?.fragment ?? ""}
@@ -76,11 +77,15 @@ export function BrevoContactForm({ id, scope, input2State, additionalFormFields,
         id ? { variables: { id, scope } } : { skip: true },
     );
 
-    const initialValues = React.useMemo<Partial<EditBrevoContactFormValues>>(() => {
+    const initialValues = React.useMemo<Partial<BaseBrevoContactFormValues>>(() => {
         let additionalInitialValues = {};
 
         if (input2State) {
-            additionalInitialValues = input2State({ email: "", redirectionUrl: "", ...data?.brevoContact });
+            additionalInitialValues = input2State({
+                email: "",
+                redirectionUrl: "",
+                ...data?.brevoContact,
+            });
         }
         return data?.brevoContact
             ? {
@@ -112,7 +117,7 @@ export function BrevoContactForm({ id, scope, input2State, additionalFormFields,
         },
     });
 
-    const handleSubmit = async (state: EditBrevoContactFormValues, form: FormApi<EditBrevoContactFormValues>, event: FinalFormSubmitEvent) => {
+    const handleSubmit = async (state: BaseBrevoContactFormValues, form: FormApi<BaseBrevoContactFormValues>, event: FinalFormSubmitEvent) => {
         if (await saveConflict.checkForConflicts()) {
             throw new Error("Conflicts detected");
         }
@@ -159,7 +164,7 @@ export function BrevoContactForm({ id, scope, input2State, additionalFormFields,
     }
 
     return (
-        <FinalForm<EditBrevoContactFormValues> apiRef={formApiRef} onSubmit={handleSubmit} mode={mode} initialValues={initialValues}>
+        <FinalForm<BaseBrevoContactFormValues> apiRef={formApiRef} onSubmit={handleSubmit} mode={mode} initialValues={initialValues}>
             {({ values }) => (
                 <EditPageLayout>
                     {saveConflict.dialogs}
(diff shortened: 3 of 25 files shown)


demo/admin/src/common/brevoModuleConfig/brevoContactsPageAttributesConfig.tsx
```diff
@@ -110,10 +108,8 @@ export const getBrevoContactConfig = (
...
             </>
         ),
-        input2State: (values?: AdditionalFormConfigInputProps) => {
+        input2State: (values: AdditionalFormConfigInputProps) => {
             return {
-                email: values?.email ?? "",
-                redirectionUrl: values?.redirectionUrl ?? "",
                 attributes: {
                     SALUTATION: values?.attributes?.SALUTATION,
                     FIRSTNAME: values?.attributes?.FIRSTNAME,
```
RainbowBunchie, line 115: Since the values are not nullable anymore, you can remove the '?' here as well

packages/admin/src/brevoContacts/form/BrevoContactForm.tsx
```diff
@@ -42,9 +42,13 @@ import {
 } from "./BrevoContactForm.gql.generated";
 
 export type EditBrevoContactFormValues = {
+    [key: string]: unknown;
+};
+
+type BaseBrevoContactFormValues = EditBrevoContactFormValues & {
+    //Umbenennen
```
RainbowBunchie, line 49: Should hits still be renamed? Or can the comment be deleted?

packages/admin/src/brevoContacts/form/BrevoContactForm.tsx
```diff
@@ -76,11 +77,15 @@ export function BrevoContactForm({ id, scope, input2State, additionalFormFields,
...
 
-    const initialValues = React.useMemo<Partial<EditBrevoContactFormValues>>(() => {
+    const initialValues = React.useMemo<Partial<BaseBrevoContactFormValues>>(() => {
         let additionalInitialValues = {};
 
         if (input2State) {
-            additionalInitialValues = input2State({ email: "", redirectionUrl: "", ...data?.brevoContact });
+            additionalInitialValues = input2State({
+                email: "",
+                redirectionUrl: "",
```
RainbowBunchie, line 86: these values should be in the `additionalInitialValues` (would rename to `baseInitalValues`). Since they are not needed in the `input2State` anymore. And the `input2state` values should be added to these `baseInitalValues`

packages/api/src/mikro-orm/migrations/Migration20240830112400.ts
```diff
@@ -0,0 +1,7 @@
+import { Migration } from "@mikro-orm/migrations";
+
+export class Migration20240830112400 extends Migration {
+    async up(): Promise<void> {
+        this.addSql('alter table "TargetGroup" add column "isTestList" boolean null;');
```
RainbowBunchie, line 5: This should not be nullable. You can set existing values to `false` since there are no existing test lists because it is a new feature.

packages/api/src/brevo-contact/brevo-contacts.service.ts
```diff
@@ -53,12 +53,31 @@ export class BrevoContactsService {
...
+    }
+
     public async getTargetGroupIdsForNewContact({
         contactAttributes,
         scope,
+        assignedContactsTargetGroupBrevoId,
     }: {
         contactAttributes?: BrevoContactAttributesInterface;
         scope?: EmailCampaignScopeInterface;
+        assignedContactsTargetGroupBrevoId?: number;
```
RainbowBunchie, line 80: This is not used, please remove

packages/api/src/brevo-contact/brevo-contacts.service.ts
```diff
@@ -53,12 +53,31 @@ export class BrevoContactsService {
...
+        attributes,
+        scope,
+    }: {
+        email: string;
+        attributes?: BrevoContactAttributesInterface;
+        scope: EmailCampaignScopeInterface;
+    }): Promise<boolean> {
+        const testTargetGroupForScope = await this.targetGroupService.createIfNotExistTestTargetGroupForScope(scope);
+
+        const brevoIds = [testTargetGroupForScope.brevoId];
```
RainbowBunchie, line 67: imo you do not need this variable, can simply pass the array like this in the brevo `createTestContact` function.

packages/api/src/brevo-contact/brevo-contact.resolver.ts
```diff
@@ -92,6 +95,39 @@ export function createBrevoContactResolver({
             return new PaginatedBrevoContacts(contacts, count, { offset, limit });
         }
 
+        @Query(() => PaginatedBrevoContacts)
+        async brevoTestContacts(@Args() { offset, limit, email, targetGroupId, scope }: BrevoContactsArgs): Promise<PaginatedBrevoContacts> {
```
RainbowBunchie, line 99: You can remove the `targetGroupId` here -> this query can only be used to query the test list. if you set a   `targetGroupId` in the args that is not the test list id this will not return anything anyways

RainbowBunchie (changes requested):

demo/admin/src/common/brevoModuleConfig/brevoContactsPageAttributesConfig.tsx
juliawegmayr: If I do so, I get the error:
_Cannot read properties of undefined (reading 'attributes')_, when I open the BrevoContactForm 🤔 

Did we change anything for those values? I think, they still can be undefined

RainbowBunchie: In that case the `?` should be readded in the function params in line 111. I do not know why this was removed 🤷🏻‍♀️

packages/api/src/mikro-orm/migrations/Migration20240830112400.ts
RainbowBunchie: Drop the default value after setting all rows to false 🙏🏻

packages/admin/src/brevoContacts/form/BrevoContactForm.tsx
RainbowBunchie: Now that I think about it: I do not think `Base` is not really the correct name for this. Since this extends the BaseFormValues (defined in `& { ...}`} with the configured values. Something like `BrevoContactFormValuesWithAttributes` would be easier to understand I think. What do you think?

juliawegmayr: Maybe `EditBrevoContactFormValuesWithAttributes`?

RainbowBunchie: 👍🏻

packages/api/src/brevo-contact/brevo-contact.resolver.ts
```diff
@@ -184,13 +210,103 @@ export function createBrevoContactResolver({
...
+        @Mutation(() => Boolean)
+        @AffectedEntity(BrevoContact)
+        async deleteBrevoTestContact(
+            @Args("id", { type: () => Int }) id: number,
+            @Args("scope", { type: () => Scope }, new DynamicDtoValidationPipe(Scope)) scope: typeof Scope,
+        ): Promise<boolean> {
+            const contact = await this.brevoContactsApiService.findContact(id, scope);
+            const where: FilterQuery<TargetGroupInterface> = { scope, isMainList: false, isTestList: true };
+            const testTargetGroup = await this.targetGroupRepository.findOne(where);
+            const contactIncludesMainList = testTargetGroup?.brevoId ? contact.listIds.includes(testTargetGroup.brevoId) : false;
```
raphaelblum, line 291: why `contactIncludesMainList`?
shouldn't that be `contactIncludesTestList` ?

packages/api/src/brevo-api/brevo-api-contact.service.ts
```diff
@@ -95,8 +110,8 @@ export class BrevoApiContactsService {
             if (!contact) return undefined;
             return contact;
         } catch (error) {
-            // Brevo throws 400 error if no contact was found
-            if (isErrorFromBrevo(error) && error.response.statusCode === 400) {
+            // Brevo throws 404 error if no contact was found
+            if (isErrorFromBrevo(error) && error.response.statusCode === 404) {
```
raphaelblum, line 114: why is this changed? In the main branch thomas changed that to catch both errors:
https://github.com/vivid-planet/comet-brevo-module/pull/94/files

juliawegmayr: I had problems, as I got a 404 error and I thought that the wrong code was given in the code. I checked it out at Brevo now, and I agree, that both codes should be handled. Thanks for fixing it and pointing it out here as well. 👍

packages/api/src/target-group/target-groups.service.ts
```diff
@@ -141,6 +139,27 @@ export class TargetGroupsService {
...
     }
 
+    public async createIfNotExistTestTargetGroupForScope(scope: EmailCampaignScopeInterface): Promise<TargetGroupInterface> {
+        const testList = await this.repository.findOne({ scope, isMainList: false, isTestList: true });
+
+        if (testList) {
+            return testList;
+        }
+
+        const title = "Test list for current scope";
```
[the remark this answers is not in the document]
RainbowBunchie: I guess just stringifying the scope would be enough here? @raphaelblum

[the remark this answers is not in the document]
juliawegmayr: Thank you :)
I didn't think about it

packages/api/src/mikro-orm/migrations/Migration20240830112400.ts
juliawegmayr, line 1: @RainbowBunchie 
When I checked out the branch, I realised, that I get an error, when creating target group lists, stating that testList was not found. I am not sure, if this is correct, but I deleted the drop default again, as it resolved all errors.

RainbowBunchie: the dropping of the default should be included in the migration. what exactly was the error here? You can also set a default value in typescript for the `isTestList`, so the default value is apparent when looking in the code. The problem with database default values is that they are really hard to trace.

packages/api/src/target-group/target-groups.service.ts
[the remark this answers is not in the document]
raphaelblum: I think it would still be better to make it more user-readable. 
But we can think of improving this in another task, since we also should update it for the main list

.changeset/chilly-dogs-draw.md
```diff
@@ -0,0 +1,8 @@
+---
+"@comet/brevo-admin": major
+"@comet/brevo-api": major
+---
+
+Adds `createBrevoTestContactsPage` for creating a test contacts page, that is indepent from the main list.
+
+Remove `email` and `redirectionUrl` from `brevoContactsPageAttributesConfig`
```
raphaelblum, line 8: Couldn't this be a minor version?

There are no breaking changes as far as I can see. It's a new feature with the test list, but it must not be used as far as I can see.

packages/api/src/mikro-orm/migrations/Migration20240830112400.ts
raphaelblum: Yes we should remove the default again. There was probably another error.  You could check out the main branch and execute the fixtures, add some data and then switch back to this branch and check if the error is still there.

We can also take a look at the error together if it occurs again

packages/api/src/brevo-contact/dto/brevo-test-contact-input.factory.ts
```diff
@@ -0,0 +1,61 @@
...
+        })
+        class BrevoTestContactInputBase implements BrevoTestContactInputInterface {
+            @IsNotEmpty()
+            @IsString()
+            @Field()
+            email: string;
+
+            @IsBoolean()
+            @Field()
+            @IsOptional()
```
raphaelblum, line 39: if possible, please rather use `IsUndefinable` or `IsNullable`

(there might be a lot of places where this is not the case yet, but when starting the project IsUndefinable and IsNullable where not yet available)

packages/admin/src/targetGroups/TargetGroupsGrid.tsx
```diff
@@ -263,12 +264,12 @@ export function TargetGroupsGrid({
...
     ];
 
-    const { filter: gqlFilter, search: gqlSearch } = muiGridFilterToGql(columns, dataGridProps.filterModel);
+    const { search: gqlSearch } = muiGridFilterToGql(columns, dataGridProps.filterModel);
 
     const { data, loading, error } = useQuery<GQLTargetGroupsGridQuery, GQLTargetGroupsGridQueryVariables>(targetGroupsQuery, {
         variables: {
             scope,
-            filter: gqlFilter,
+            filter: { isTestList: { equal: false || null } },
```
raphaelblum, line 272: isn't this:
`filter: { isTestList: { equal: false || null } },`

the same as this:
`filter: { isTestList: { equal: null } },`

.changeset/chilly-dogs-draw.md
juliawegmayr: True, I initially thought, that the change in the `brevoContactsPageAttributesConfig` breaks the contacts page and the create/update functions. But that's not true, so minor should be fine 👍

packages/api/src/mikro-orm/migrations/Migration20240830112400.ts
juliawegmayr: I'm so sorry, I tested it again and it worked perfectly fine. I guess, I had the problems because of switching branches. 
Sorry for the confusion and thanks for the clarification!

packages/admin/src/targetGroups/TargetGroupsGrid.tsx
RainbowBunchie: `isTestList` is not nullable though

.changeset/chilly-dogs-draw.md
RainbowBunchie: Is the change in `EditBrevoContactFormValues` not a breaking change? @raphaelblum @juliawegmayr 
just to be sure, i thought this is breaking since before the email and redirectionUrl had to be included in the `createBrevoContactPage` and now they should not be present there. Or am I wrong here?
 
see changes in `demo/admin/src/common/brevoModuleConfig/brevoContactsPageAttributesConfig.tsx`

```
export type EditBrevoContactFormValues = {
    [key: string]: unknown;
    // email: string; got removed 
    // redirectionUrl: string;  got removed
};
```

raphaelblum: It is still possible to pass the email and redirectionUrl, or is it not?

packages/api/src/brevo-contact/brevo-contact.resolver.ts
```diff
@@ -184,13 +210,103 @@ export function createBrevoContactResolver({
...
+                    {
+                        listIds: linkListIds,
+                        unlinkListIds: [testListId],
+                    },
+                    scope,
+                );
+                return true;
+            } else {
+                return this.brevoContactsApiService.deleteContact(id, scope);
+            }
```
raphaelblum, line 309: What exactly should happen here?

**Am I right with that:**
When the contact is only in the testlist, and nowhere else, it should be deleted.
Otherwise it should just be unlinked

**Currently it works like this:**
 - When there is a testTargetGroup AND the contact is in the list, it will be unlinked, otherwise it will be deleted? Does this make sense?

.changeset/chilly-dogs-draw.md
juliawegmayr: As long as the test contacts are not used, it is still possible to pass email and redirectionUrl.
I also tested it by using the initial config, like it is now on the main branch.

RainbowBunchie: I thought this would result in an error. If not, then there is no problem 👍🏻  sorry for the confusion

raphaelblum: > As long as the test contacts are not used, it is still possible to pass email and redirectionUrl.

Okay good then it is non-breaking. The target branch can be switched to main as @RainbowBunchie said.

packages/admin/src/targetGroups/TargetGroupsGrid.tsx
juliawegmayr: MainList is not nullable. 
I checked the code again and adapted it, so that isMainList and isTestList are both not nullable.

packages/api/src/brevo-contact/brevo-contact.resolver.ts
juliawegmayr: Yes, you are absolutely right! 
Now I check, if the contact is in the MainList and if he is, the test list is unlinked. Otherwise deleted

packages/admin/src/targetGroups/TargetGroupsGrid.tsx
juliawegmayr: c9a0e639675b628a572cb301f2df744e1cee88d8

raphaelblum (dismissed):

raphaelblum (approved):

demo/admin/src/common/brevoModuleConfig/brevoContactsPageAttributesConfig.tsx
```diff
@@ -70,9 +70,7 @@ export const getBrevoContactConfig = (
         name: string;
     };
     input2State: (values?: AdditionalFormConfigInputProps) => {
-        email: string;
-        redirectionUrl: string;
-        attributes: { BRANCH?: Array<GQLBrevoContactBranch>; SALUTATION?: GQLBrevoContactSalutation; FIRSTNAME?: string; LASTNAME?: string };
+        attributes: { SALUTATION?: GQLBrevoContactSalutation; FIRSTNAME?: string; LASTNAME?: string };
```
RainbowBunchie, line 73: Why did the branch get removed here?

packages/admin/src/targetGroups/TargetGroupsGrid.tsx
```diff
@@ -50,6 +50,7 @@ const targetGroupsFragment = gql`
         totalSubscribers
         totalContactsBlocked
         isMainList
+        isTestList
```
RainbowBunchie, line 53: This is not used in the frontend, can this be removed

RainbowBunchie: will be false for every target group anyways since it gets filtered in the api

RainbowBunchie (changes requested):

demo/admin/src/common/brevoModuleConfig/brevoContactsPageAttributesConfig.tsx
juliawegmayr: c9a0e639675b628a572cb301f2df744e1cee88d8

packages/admin/src/targetGroups/TargetGroupsGrid.tsx
juliawegmayr: 51c898ce3c997e11ef38c54df1ba856b202a056c

packages/admin/src/targetGroups/TargetGroupsGrid.tsx
```diff
@@ -263,12 +263,11 @@ export function TargetGroupsGrid({
         },
     ];
 
-    const { filter: gqlFilter, search: gqlSearch } = muiGridFilterToGql(columns, dataGridProps.filterModel);
+    const { search: gqlSearch } = muiGridFilterToGql(columns, dataGridProps.filterModel);
```
RainbowBunchie, line 266: why did the filters get removed here?
the targetgroup grid has filters right?

RainbowBunchie:
@juliawegmayr I just checked out the branch and the test list shows up in the target groups 
Can you please take a look at this? 
Screenshot 2024-10-25 at 12 55 17

RainbowBunchie (changes requested):

packages/admin/src/targetGroups/TargetGroupsGrid.tsx
juliawegmayr: Initially I wanted to filter for the test contact list here, not in the api. After undoing my changes, I forgot to add the filters here again. I'm sorry! 
I corrected it here: fd436ca3cf5abca28f238d5af16f035623c85606

juliawegmayr:
> @juliawegmayr I just checked out the branch and the test list shows up in the target groups Can you please take a look at this? Screenshot 2024-10-25 at 12 55 17

Fixed here: 205b281f06e28621620ef15f4c9707978add1d13

RainbowBunchie:
> > @juliawegmayr I just checked out the branch and the test list shows up in the target groups Can you please take a look at this? Screenshot 2024-10-25 at 12 55 17
> 
> Fixed here: [205b281](https://github.com/vivid-planet/comet-brevo-module/commit/205b281f06e28621620ef15f4c9707978add1d13)

@juliawegmayr Shouldn't this work with the "normal" filters above? in "options.filter". Why do you need the custom filter logic here?

juliawegmayr:
> > > @juliawegmayr I just checked out the branch and the test list shows up in the target groups Can you please take a look at this? Screenshot 2024-10-25 at 12 55 17
> > 
> > 
> > Fixed here: [205b281](https://github.com/vivid-planet/comet-brevo-module/commit/205b281f06e28621620ef15f4c9707978add1d13)
> 
> @juliawegmayr Shouldn't this work with the "normal" filters above? in "options.filter". Why do you need the custom filter logic here?

True! I removed my changes again: ebfe653f7989bb2afa644f54a2f36bd0a76f7873
Sorry for my missunderstandings.

raphaelblum (dismissed):

RainbowBunchie:
@juliawegmayr I just checked this out and I can still see the test group in the target groups. Did you test this? 
I would suggest you set `isTestList` to `false` in the targetGroups query in the api.

juliawegmayr:
> @juliawegmayr I just checked this out and I can still see the test group in the target groups. Did you test this? I would suggest you set `isTestList` to `false` in the targetGroups query in the api.


I guess I didn't restart the api while I tested it and of course it seemed to work then. Totally my bad. I updated the code here:

4a9af5d1a41fcee8cf614f32daba016238640eaa

raphaelblum (approved):

PR #81, opened 2024-09-02
Update to COMET v7

thomasdax98:
Update from COMET v6 to v7

**Breaking Changes**:

-   Now requires >= v7.5.0 for `@comet` packages
-   Now requires >= 5.9.8 for `@mikro-orm` packages (except `@mikro-orm/nestjs` where >= 5.2.3 is required)

No changes to the public API were necessary

### Note

- The Demo API currently contains a workaround for a COMET bug that can be removed in the future (https://github.com/vivid-planet/comet-brevo-module/commit/9bce002dcf656fad122f1fd11e6ec5220ce16a81)
- In the campaign it was necessary to migrate from the deprecated `mjml-react@2` to the maintained fork `@faire/mjml-react@3`. This was necessary because v2 doesn't support React 18 (but is a good thing anyway). Since [v3](https://github.com/Faire/mjml-react/releases/tag/v3.0.0) doesn't have impactful breaking changes, the migration shouldn't be a problem.

---

In the beginning, I split the changes into small commits. So it may make sense to review commit by commit. However, in the end it got a bit messy 😅 

---

https://vivid-planet.atlassian.net/browse/COM-1054


demo/api/.prettierignore
```diff
@@ -1,3 +1,7 @@
 dist/
 block-meta.json
-uploads/
\ No newline at end of file
+<<<<<<< HEAD
+uploads
+=======
+uploads/
+>>>>>>> next
```
raphaelblum, line 7: here are merge conflict markers

demo/api/src/app.module.ts
```diff
@@ -137,10 +132,10 @@ export class AppModule {
...
                 MenusModule,
                 DependenciesModule,
-                PublicUploadModule.register({
+                FileUploadsModule.register({
                     acceptedMimeTypes: ["text/csv"],
-                    maxFileSize: config.publicUploads.maxFileSize,
-                    directory: `${config.blob.storageDirectoryPrefix}-public-uploads`,
+                    maxFileSize: config.fileUploads.maxFileSize,
+                    directory: `${config.blob.storageDirectoryPrefix}-file-uploads`,
                 }),
```
raphaelblum, line 139: @RainbowBunchie With this the new fileUpload module is here. Could you think of any problems with this in the brevo module?

And we could now use the comet file upload handler instead of making it custom, right?

demo/campaign/src/documents/pages/blocks/SeoBlock.tsx
```diff
@@ -14,30 +14,24 @@ export const SeoBlock: React.FunctionComponent<SeoBlockProps> = ({
 }) => {
     const usedHtmlTitle = htmlTitle && htmlTitle != "" ? htmlTitle : title;
     return (
-        <>
-            <Head>
-                <title>{usedHtmlTitle}</title>
+        <Head>
+            <title>{usedHtmlTitle}</title>
```
raphaelblum, line 18: For future PR's:
This block shouldn't be needed here. SEO does not make sense in emails. I'll create a task

raphaelblum (approved):

demo/api/.prettierignore
thomasdax98: 5105e4aea7c638b92d1ad45434c3d6053ff62eda

demo/api/src/app.module.ts
thomasdax98: I just realised that I didn't test the upload. And I forgot to change the URL: https://github.com/vivid-planet/comet-brevo-module/blob/40ba7d342f242cbceff33a453fc7e5a77337a496/packages/admin/src/common/contactImport/useContactImportFromCsv.tsx#L70

I still have to do that.

Apart from that:

For the most part, PublicUpload was just renamed to FileUpload. So there shouldn't be any hard problems. 

I considered if changing the directory name would cause problems. But since all uploaded files are deleted once the import is complete, this shouldn't cause problems. Also it's a config in the project so it can be decided on a project-basis. For the demo I think it's right to rename it.

> And we could now use the comet file upload handler instead of making it custom, right?

The upload endpoint is already provided by COMET. However, the get file and delete file logic must stay like this because we don't have methods for this in COMET yet

RainbowBunchie: If the module still works the same I guess there should not be any problems here.

thomasdax98: I changed the URL, tested it and it works: 9fa549a18b4da47dff911097f59b294369af0a9c

RainbowBunchie (approved):

PR #97, opened 2024-09-20
COM-964: support multiselect in contact attributes

juliawegmayr:
- Add demo contact attribute (branch) allowing to select multiple fields.
- Add demo filter for branch for testing.

Logic for assigning contacts to filters:
- If multiple filter are set in a target group, the contact has to have all those attributes. (e.g. SALUTATION: "Female", BRANCH: "Marketing" -> contact needs both attributes to be assigned.
- If a filter value is an array, one matching  value is required for assigning a contact. (e.g. BRANCH: ["Marketing", "Products"] -> contact only needs one of the values.

Contact Attributes:
Screenshot 2024-09-20 at 09 31 48

Filter:
Screenshot 2024-09-25 at 09 36 38

### DIFF
--- a/.changeset/fifty-radios-wonder.md
+++ b/.changeset/fifty-radios-wonder.md
@@ -1,0 +1,7 @@
+---
+"@comet/brevo-api": minor
+---
+
+Add check for arrays in `checkIfContactIsInTargetGroup` function, to check if at least one `contactAttribute` is included in the filter.
+
+Pass `scope` to `getTargetGroupIdsForExistingContact`.
--- a/.changeset/fifty-radios-wonder.md
+++ b/.changeset/fifty-radios-wonder.md
@@ -3,5 +3,3 @@
 ---
 
 Add check for arrays in `checkIfContactIsInTargetGroup` function, to check if at least one `contactAttribute` is included in the filter.
-
-Pass `scope` to `getTargetGroupIdsForExistingContact`.
--- a/.changeset/fifty-radios-wonder.md
+++ b/.changeset/fifty-radios-wonder.md
@@ -1,5 +1,5 @@
 ---
-"@comet/brevo-api": minor
+"@comet/brevo-api": patch
 ---
 
 Add check for arrays in `checkIfContactIsInTargetGroup` function, to check if at least one `contactAttribute` is included in the filter.
--- a/packages/api/src/brevo-contact/brevo-contact.resolver.ts
+++ b/packages/api/src/brevo-contact/brevo-contact.resolver.ts
@@ -143,7 +143,6 @@ export function createBrevoContactResolver({
 
             const updatedNonMainListIds = await this.brevoContactsService.getTargetGroupIdsForExistingContact({
                 contact,
-                scope,
             });
 
             // update contact again with updated list ids depending on new attributes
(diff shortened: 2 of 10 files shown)


packages/api/src/target-group/target-groups.service.ts
```diff
@@ -37,16 +37,23 @@ export class TargetGroupsService {
     ): boolean {
         if (filters && contactAttributes) {
             for (const [key, value] of Object.entries(filters)) {
-                if (!value) continue;
+                if (!value || value.length === 0) continue;
+
+                if (Array.isArray(contactAttributes[key])) {
+                    for (const attribute of contactAttributes[key]) {
+                        if (value.includes(attribute)) {
+                            return true;
```
RainbowBunchie, line 45: You shouldn't return true in this case. Then the other attributes that come after the array will not be checked anymore. 
Example filters (note the order of the filters):
```
BRANCH -> [NEWS] 
SALUTATION -> FEMALE 
```
This would return true for contact since you instantly return true when checking the branch array (which is wrong)
```
BRANCH -> NEWS 
SALUTATION -> MALE
``` 
you should only skip the remaining attributes in the for loop (line 43) in this case.

packages/api/src/brevo-contact/brevo-contacts.service.ts
```diff
@@ -128,10 +128,6 @@ export class BrevoContactsService {
                 if (contactIsInTargetGroupByAttributes) {
                     targetGroupIds.push(targetGroup.brevoId);
                 }
-
-                if (targetGroup.assignedContactsTargetGroupBrevoId && contact?.listIds.includes(targetGroup.assignedContactsTargetGroupBrevoId)) {
-                    targetGroupIds.push(targetGroup.brevoId, targetGroup.assignedContactsTargetGroupBrevoId);
-                }
```
RainbowBunchie, line 134: Why did you remove this? The manually assigned contact groups do not get added anymore if you remove this

RainbowBunchie (changes requested):

RainbowBunchie (changes requested):

packages/api/src/brevo-contact/brevo-contact.resolver.ts
```diff
@@ -143,6 +143,7 @@ export function createBrevoContactResolver({
 
             const updatedNonMainListIds = await this.brevoContactsService.getTargetGroupIdsForExistingContact({
                 contact,
+                scope,
```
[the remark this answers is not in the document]
juliawegmayr: I fixed it again, by removing the scope from the where args as it is not neccessary here

packages/api/src/brevo-contact/brevo-contacts.service.ts
```diff
@@ -112,7 +112,7 @@ export class BrevoContactsService {
         let totalCount = 0;
         const targetGroupIds: number[] = [];
         const limit = 50;
-        const where = { isMainList: false, scope };
+        const where = { isMainList: false };
```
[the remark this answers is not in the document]
juliawegmayr: Yes, removing is possible. The function is only used in the updateBrevoContact, where the contact is already checked for the scope, so that it's not neccessary here anymore.

.changeset/fifty-radios-wonder.md
```diff
@@ -0,0 +1,5 @@
+---
+"@comet/brevo-api": minor
```
raphaelblum, line 2: I would say this is a patch. I differentiate it like this:

patch: bugfixes, security updates or small changes
minor: New Features like adding test list functionality
major: Breaking Changes, big new Features that greatly change the use of the module and so on

juliawegmayr: Thank you for the explanation! :)

RainbowBunchie (approved):

raphaelblum (approved):

PR #102, opened 2024-09-30
COM-1062 Add validation and do not save all contacts into memory when importing

raphaelblum:
COM-1062

Empty File:

https://github.com/user-attachments/assets/62c57e29-4e06-4e52-8e0e-5bb4ba69c786

File with only wrong rows (4 wrong and 1 empty --> empty is removed automatically):
(in the video there is column(s). I changed it later to row(s)  see last video)
https://github.com/user-attachments/assets/5931a792-b393-4bf2-b1e9-3e44efa36b07

5 correct ones:

https://github.com/user-attachments/assets/200cf026-a58a-4a63-9597-9c2f61dace6f

2 invalid, 3 correct:

https://github.com/user-attachments/assets/54d7491a-11ae-4e37-974b-a6ce4cda43e7


packages/api/src/brevo-contact/brevo-contact-import.controller.ts
```diff
@@ -33,23 +35,27 @@ export function createBrevoContactImportController({ Scope }: { Scope: Type<Emai
...
-                throw new CometValidationException("Too many contacts in file. Currently we only support 100 contacts at once.");
-            }
 
             let parsedListIds = undefined;
             if (listIds) parsedListIds = JSON.parse(listIds) as string[];
 
             const targetGroups = await this.targetGroupRepository.find({ id: { $in: parsedListIds } });
 
-            await this.brevoContactImportService.importContactsFromCsv(content, parsedScope, redirectUrl, targetGroups);
+            const stream = Readable.from(file.buffer);
```
raphaelblum, line 51: I have changed this to streaming, otherwise everything would be loaded into the memory.

The disadvantage is that I can't simply say that if the file has more than 100 contacts, it won't be treated at all.

But I think we should fix the problem that only 100 contacts can be imported via the admin anyway, otherwise the feature is probably a bit pointless.

packages/api/src/brevo-contact/brevo-contact-import.service.ts
```diff
@@ -11,13 +12,29 @@ import { TargetGroupInterface } from "../target-group/entity/target-group-entity
...
     @IsEmail()
     @IsNotEmpty()
     email: string;
 
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
-    [key: string]: any;
+    [key: string]: string;
+}
+
+export interface CSVImportInformation {
```
RainbowBunchie, line 24:
```
export interface CsvImportInformation {
```

packages/api/src/brevo-contact/brevo-contact-import.service.ts
```diff
@@ -11,13 +12,29 @@ import { TargetGroupInterface } from "../target-group/entity/target-group-entity
...
+    created: number;
+    updated: number;
+    failed: number;
+    failedColumns: unknown[];
+    errorMessage?: string;
+}
+
+interface ImportContactsFromScvParams {
```
RainbowBunchie, line 32:
```
interface ImportContactsFromCsvParams {
```

packages/api/src/brevo-contact/brevo-contact-import.service.ts
```diff
@@ -29,114 +46,151 @@ export class BrevoContactImportService {
...
+            try {
+                brevoContact = await this.brevoApiContactsService.findContact(contact.email, scope);
+            } catch (error) {
+                // Brevo throws 404 error if no contact was found
+                if (!isErrorFromBrevo(error)) {
+                    throw error;
+                }
+                if (error.response.statusCode !== 404) {
+                    throw error;
+                }
```
RainbowBunchie, line 127: The comment should be above the second if i think

packages/api/src/brevo-contact/brevo-contact-import.service.ts
```diff
@@ -29,114 +46,151 @@ export class BrevoContactImportService {
...
+            }
+            if (brevoContact && !brevoContact.emailBlacklisted) {
+                const mainTargetGroupForScope = await this.targetGroupsService.createIfNotExistMainTargetGroupForScope(scope);
+
+                const updatedBrevoContact = await this.brevoApiContactsService.updateContact(
+                    brevoContact.id,
+                    { ...contact, listIds: [mainTargetGroupForScope.brevoId, ...targetGroupBrevoIds, ...brevoContact.listIds] },
```
RainbowBunchie, line 134: The targetGroupBrevoIds are only the manuallyAssigned targetGroupIds if I understand it correctly. Does this also include the contact in the "main" list of the target group? Since the manuallyAssigned group is only a helper target group? Does the contact end up in the recipients this way?

packages/api/src/brevo-contact/brevo-contact-import.controller.ts
RainbowBunchie: I agree, the admin import should be handled differently so more than 100 contacts can be imported. So this temporary solution is fine imo

packages/admin/src/common/contactImport/useContactImportFromCsv.tsx
```diff
@@ -157,3 +228,9 @@ const ContactImportComponent = ({ scope, targetGroupId, fileInputRef, refetchQue
         </>
     );
 };
+
+const CSVDownloadLink = styled("span")`
```
RainbowBunchie, line 232:
```
const CsvDownloadLink = styled("span")`
```

packages/api/src/brevo-contact/brevo-contact-import.service.ts
raphaelblum: yes, thanks.

done in: 7254d7fd4219199315a3b0834f01a6a6537c6304

packages/api/src/brevo-contact/brevo-contact-import.service.ts
raphaelblum: done in: 7254d7fd4219199315a3b0834f01a6a6537c6304

packages/api/src/brevo-contact/brevo-contact-import.service.ts
raphaelblum: done in: 7254d7fd4219199315a3b0834f01a6a6537c6304

packages/admin/src/common/contactImport/useContactImportFromCsv.tsx
raphaelblum: done in: 7254d7fd4219199315a3b0834f01a6a6537c6304

packages/api/src/brevo-contact/brevo-contact-import.service.ts
raphaelblum: I did not specifically check this code because I didn't change anything there. I'll check it now

raphaelblum: Thank you. This did not work correctly as you said, fixed in: 24a42a930ddf8939027edf0e06151e15aae1ba40

raphaelblum: 72332f98f4f7b1a8b0b8d390fca671caedd55607

added validation for main lists and for the correct scope: 72332f98f4f7b1a8b0b8d390fca671caedd55607

manually assigned target group ids are defined here:
https://github.com/vivid-planet/comet-brevo-module/blob/main/packages/api/src/target-group/entity/target-group-entity.factory.ts#L78

and wouldn't be found when looking for targetGroups by brevoId.

test list ids are not yet merged

RainbowBunchie: Thanks for checking! 
Looking good now

RainbowBunchie (approved):

PR #104, opened 2024-10-03
COM-1135: Improve Brevo Error Messages

juliawegmayr:
Errors from the Brevo API weren't explicitly handled. This error report wasn't helpful and devs would have to view the API logs to understand the actual error. To improve this, we now wrap all Brevo API operations in a try/catch block and handle errors explicitly.

## Screenshots/screencasts
Example: The entered redirection URL was not a valid URL:
Before:
Screenshot 2024-10-03 at 07 56 01

After:
Screenshot 2024-10-03 at 07 57 21

## Changeset

[COM-1135](https://vivid-planet.atlassian.net/browse/COM-1135)

[COM-1135]: https://vivid-planet.atlassian.net/browse/COM-1135?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/api/src/brevo-api/brevo-api.utils.ts
+++ b/packages/api/src/brevo-api/brevo-api.utils.ts
@@ -1,5 +1,26 @@
+import { ErrorModel } from "@getbrevo/brevo";
 import * as http from "http";
 
-export function isErrorFromBrevo(error: unknown): error is { response: http.IncomingMessage } {
-    return typeof error === "object" && error !== null && "response" in error && error.response instanceof http.IncomingMessage;
+export function isErrorFromBrevo(error: unknown): error is { response: http.IncomingMessage; body: ErrorModel } {
+    return (
+        typeof error === "object" &&
+        error !== null &&
+        "response" in error &&
+        error.response instanceof http.IncomingMessage &&
+        "body" in error &&
+        typeof error.body === "object" &&
+        error.body !== null &&
+        "code" in error.body &&
+        typeof error.body.code === "string" &&
+        "message" in error.body &&
+        typeof error.body.message === "string"
+    );
+}
+
+export function handleBrevoError(error: unknown): never {
+    if (isErrorFromBrevo(error)) {
+        throw new Error(error.body.message);
+    } else {
+        throw new Error();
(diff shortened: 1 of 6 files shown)


raphaelblum (dismissed):

packages/api/src/brevo-api/brevo-api.utils.ts
```diff
@@ -1,5 +1,26 @@
...
+        "message" in error.body &&
+        typeof error.body.message === "string"
+    );
+}
+
+export function handleBrevoError(error: unknown): never {
+    if (isErrorFromBrevo(error)) {
+        throw new Error(error.body.message);
+    } else {
+        throw new Error();
```
RainbowBunchie, line 24: Shouldn't this re-throw the inital error? 
Eg.: https://github.com/vivid-planet/comet-brevo-module/pull/104/files#diff-f7b10559b572b094baf502b0beda3119a997258f6396a41344f75de5bddbe41dR148

packages/api/src/target-group/target-groups.service.ts
```diff
@@ -66,53 +67,62 @@ export class TargetGroupsService {
...
-                    scope,
-                );
-            }
-        } while (offset < totalCount);
-
-        return true;
+            return true;
+        } catch (error) {
+            handleBrevoError(error);
+            throw error;
```
RainbowBunchie, line 124: Why is this needed? handleBrevo throws an error in any case, so this line can never be reached right?

juliawegmayr: 88bcc36438b1f9ebc033777055f07ec9f2b46abe

packages/api/src/brevo-api/brevo-api.utils.ts
juliawegmayr: 806b2daba08c0b5f5fb1bcd57d9da40c16d2fca5

RainbowBunchie (approved):

PR #105, opened 2024-10-08
Adapt import to temporarily store file till processed

RainbowBunchie:
## Description

The files for the brevo contact import now get temporarily stored in the public uploads until the import is concluded.
This change prepares for future imports to be handled in a separate job, allowing more than 100 contacts to be imported without exhausting api resources or blocking the event loop.

It is now necessary to import the `PublicUploadsModule` in the project's `AppModule` and configure it to accept csv files.

```ts
        PublicUploadModule.register({
            acceptedMimeTypes: ["text/csv"],
            maxFileSize: config.publicUploads.maxFileSize,
            directory: `${config.blob.storageDirectoryPrefix}-public-uploads`,
        })
```

## Example

## Changeset

## Open TODOs/questions

COM-1137

## Further information

In the upcoming merge requests a separate job will handle the import


raphaelblum:
We also need a job that deletes all files after a while (but only the files for the brevo module)

Or do you know if the publicApi Module is deleting files automatically after a while? I don't think so

raphaelblum (approved):

PR #108, opened 2024-10-11
COM-1135: Fix brevo error handling

RainbowBunchie:
## Description

Fix import failing due to error in brevo handeling. 
The import was causing an error because of nested try catch blocks and adapting the brevo error message. Now the `findContact` function returns `undefined` if no brevo contact was fround instead of an error. 

## Changeset

[x ] I have verified if my change requires a changeset

COM-1135

### DIFF
--- a/packages/api/src/brevo-api/brevo-api-contact.service.ts
+++ b/packages/api/src/brevo-api/brevo-api-contact.service.ts
@@ -108,7 +108,7 @@ export class BrevoApiContactsService {
         }
     }
 
-    public async findContact(idOrEmail: string | number, scope: EmailCampaignScopeInterface): Promise<BrevoContactInterface | undefined> {
+    public async findContact(idOrEmail: string | number, scope: EmailCampaignScopeInterface): Promise<BrevoContactInterface | null> {
         try {
             const idAsString = String(idOrEmail); // brevo expects a string, because it can be an email or the id
             const { body } = await this.getContactsApi(scope).getContactInfo(idAsString);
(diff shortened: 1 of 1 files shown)


packages/api/src/brevo-api/brevo-api-contact.service.ts
```diff
@@ -101,13 +108,18 @@ export class BrevoApiContactsService {
         }
     }
 
-    public async findContact(idOrEmail: string | number, scope: EmailCampaignScopeInterface): Promise<BrevoContactInterface> {
+    public async findContact(idOrEmail: string | number, scope: EmailCampaignScopeInterface): Promise<BrevoContactInterface | undefined> {
```
raphaelblum, line 111: Optional:
In my opinion null would be more appropriate in this case. It's how nest also handles it when you check in the database if an entry exists. It return null.

Undefined is in my opinion more like "unassigned", in this case "null" would mean the lookup for the contact was made, but there was nothing.

But maybe that's just me and we don't have a best practice as far as I know, so both is fine for me.

raphaelblum (approved):

packages/api/src/brevo-api/brevo-api-contact.service.ts
RainbowBunchie: [44f4fbb](https://github.com/vivid-planet/comet-brevo-module/pull/108/commits/44f4fbb035e3503331a86614a2207463105bd03e)

raphaelblum (approved):

PR #124, opened 2024-11-14
Remove blocked subscribers from target groups query

raphaelblum:
## Description

Brevo does not deliver totalBlacklisted anymore in the list query, see: https://developers.brevo.com/reference/getlists-1

Screenshot 2024-11-14 at 07 34 51

Loading the totalBlacklisted count per ResolveField does not work either, because the contacts api is limited:
https://developers.brevo.com/docs/api-limits#general-rate-limiting

We can only make 20 contacts query per second, so when a scope has a lot of target groups, the query just fails.

I removed the blacklisted count for now. Should we decide that we still need it, we will need to sync the data regularly and add it back again.

## Screenshots/screencasts

Screenshot 2024-11-14 at 07 38 32

## Changeset

### DIFF
--- a/.changeset/short-feet-sing.md
+++ b/.changeset/short-feet-sing.md
@@ -1,0 +1,6 @@
+---
+"@comet/brevo-admin": patch
+"@comet/brevo-api": patch
+---
+
+Remove the field `Total contacts blocked` in the Target Groups query, because it is not delivered in the list request in Brevo anymore.
--- a/.changeset/short-feet-sing.md
+++ b/.changeset/short-feet-sing.md
@@ -3,4 +3,4 @@
 "@comet/brevo-api": patch
 ---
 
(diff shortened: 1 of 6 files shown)


.changeset/short-feet-sing.md
```diff
@@ -0,0 +1,6 @@
+---
+"@comet/brevo-admin": patch
+"@comet/brevo-api": patch
+---
+
+Remove the field `Total contacts blocked` in the Target Groups query, because it is not delivered in the list request in Brevo anymore.
```
johnnyomair, line 6:
```suggestion
Remove the `totalContactsBlocked` field from the `TargetGroup` type, because it is not delivered in the list request in Brevo anymore.
```

johnnyomair (approved):

thomasdax98 (approved):

PR #159, opened 2024-11-19
Add folder id to config

juliawegmayr:
## Description

Add folderId to BrevoConfig to enable overwriting the default folderId (=1) in case the folder was deleted. 

## Screenshots/screencasts

Screenshot 2024-11-19 at 13 58 30
Screenshot 2024-11-19 at 13 59 12

## Changeset

[COM-1280](https://vivid-planet.atlassian.net/browse/COM-1280)

[COM-1280]: https://vivid-planet.atlassian.net/browse/COM-1280?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/demo/api/schema.gql
+++ b/demo/api/schema.gql
@@ -460,7 +460,7 @@ type BrevoConfig implements DocumentInterface {
   updatedAt: DateTime!
   senderMail: String!
   senderName: String!
-  folderId: Int
+  folderId: Int!
   createdAt: DateTime!
   scope: EmailCampaignContentScope!
 }
--- a/demo/api/schema.gql
+++ b/demo/api/schema.gql
@@ -460,6 +460,7 @@ type BrevoConfig implements DocumentInterface {
   updatedAt: DateTime!
   senderMail: String!
   senderName: String!
+  folderId: Int
   createdAt: DateTime!
   scope: EmailCampaignContentScope!
 }
@@ -997,9 +998,11 @@ input TargetGroupUpdateInput {
 input BrevoConfigInput {
   senderMail: String!
   senderName: String!
+  folderId: Int!
 }
 
 input BrevoConfigUpdateInput {
   senderMail: String
   senderName: String
+  folderId: Int
 }
--- a/packages/admin/src/brevoConfiguration/BrevoConfigForm.tsx
+++ b/packages/admin/src/brevoConfiguration/BrevoConfigForm.tsx
@@ -7,6 +7,7 @@ import {
     FinalFormSubmitEvent,
     Loading,
     MainContent,
+    TextField,
     Toolbar,
     ToolbarActions,
     ToolbarFillSpace,
@@ -76,9 +77,10 @@ export function BrevoConfigForm({ scope }: FormProps): React.ReactElement {
                       value: sender.id,
                       label: `${sender.name} (${sender.email})`,
                   },
+                  folderId: data?.brevoConfig?.folderId,
               }
             : {};
-    }, [data?.brevoConfig?.senderMail, data?.brevoConfig?.senderName, sendersData?.senders]);
+    }, [data?.brevoConfig?.folderId, data?.brevoConfig?.senderMail, data?.brevoConfig?.senderName, sendersData?.senders]);
 
     const saveConflict = useFormSaveConflict({
         checkConflict: async () => {
@@ -117,6 +119,7 @@ export function BrevoConfigForm({ scope }: FormProps): React.ReactElement {
         const output = {
             senderName: sender?.name,
             senderMail: sender?.email,
+            folderId: data?.brevoConfig?.folderId ?? 1,
         };
 
         if (mode === "edit") {
@@ -172,6 +175,11 @@ export function BrevoConfigForm({ scope }: FormProps): React.ReactElement {
                                 options={senderOptions}
                                 name="sender"
                                 label={<FormattedMessage id="cometBrevoModule.brevoConfig.sender" defaultMessage="Sender" />}
+                                fullWidth
+                            />
+                            <TextField
+                                name="folderId"
+                                label={<FormattedMessage id="cometBrevoModule.brevoConfig.folderId" defaultMessage="Folder ID" />}
                                 fullWidth
                             />
                         </MainContent>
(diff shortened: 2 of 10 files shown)


juliawegmayr:
> Could you set it as 1 by default in the admin? (when you first open the form without data) In almost all cases this will remain 1.
> 
> I think a small info icon with a text describing that this should almost always be 1 as long as you had not configured anything special here would also be great

Sure! Thanks for the input!
Add info icon: 4f89d4d61954aafb5a5cd22ae8855d79a70243cd
Add defaultValue to the form field: 537ef29e2ed559b0b8e342ff79ac28b72ce2356e

packages/api/src/mikro-orm/migrations/Migration20241119101706.ts
```diff
@@ -0,0 +1,7 @@
+import { Migration } from "@mikro-orm/migrations";
+
+export class Migration20241119101706 extends Migration {
+    async up(): Promise<void> {
+        this.addSql('alter table "BrevoConfig" add column "folderId" int;');
```
RainbowBunchie, line 5: Is folder id required or optional - in the db it is required but in the entity it is nullable?

packages/admin/src/brevoConfiguration/BrevoConfigForm.tsx
```diff
@@ -174,6 +180,37 @@ export function BrevoConfigForm({ scope }: FormProps): React.ReactElement {
...
+                                validate={(value) => {
+                                    if (typeof value !== "number") {
+                                        return (
+                                            <FormattedMessage
+                                                id="cometBrevoModule.brevoConfig.folderId.validation"
+                                                defaultMessage="Please enter a number"
+                                            />
+                                        );
+                                    }
+                                }}
```
RainbowBunchie, line 212: I would move the validation in it's own function (outside the JSX) for better readability

RainbowBunchie:
> Could you set it as 1 by default in the admin? (when you first open the form without data) In almost all cases this will remain 1.
> 
> I think a small info icon with a text describing that this should almost always be 1 as long as you had not configured anything special here would also be great

@raphaelblum would you also save the default folder in the database and make the field required? 🤔 I think it is misleading to set the default to 1 but the field is also nullable and if not set "1" is used per default anyways

packages/admin/src/brevoConfiguration/BrevoConfigForm.tsx
juliawegmayr: Done here: 9cfd6671ad412ef23e77650c2cb8376ef1d676f7

packages/api/src/mikro-orm/migrations/Migration20241119101706.ts
juliawegmayr: I set it to required, I guess that makes more sense, as we set it to 1 anyway: c48c00f7a092a0722cf5db4ab86ef160d2a4066f

packages/api/src/mikro-orm/migrations/Migration20241119101706.ts
```diff
@@ -0,0 +1,7 @@
+import { Migration } from "@mikro-orm/migrations";
+
+export class Migration20241119101706 extends Migration {
+    async up(): Promise<void> {
+        this.addSql('alter table "BrevoConfig" add column "folderId" int not null;');
```
raphaelblum, line 5: If a brevoconfig has already been added here, it will crash. All existing data must be set to 1 here.

packages/admin/src/brevoConfiguration/BrevoConfigForm.tsx
```diff
@@ -149,6 +155,12 @@ export function BrevoConfigForm({ scope }: FormProps): React.ReactElement {
         return <Loading behavior="fillPageHeight" />;
     }
 
+    const validateFolderId = (value: number) => {
+        if (typeof value !== "number") {
+            return <FormattedMessage id="cometBrevoModule.brevoConfig.folderId.validation" defaultMessage="Please enter a number" />;
+        }
+    };
```
RainbowBunchie, line 162: Is this validation needed? Since you use a NumberField, you cannot enter other characters anyways right?

packages/api/src/mikro-orm/migrations/Migration20241119101706.ts
juliawegmayr: a1a8d60be6929e0a8436c6a8dd1958601dc8afb5

raphaelblum (dismissed):

packages/api/src/mikro-orm/migrations/Migration20241119101706.ts
```diff
@@ -0,0 +1,7 @@
+import { Migration } from "@mikro-orm/migrations";
+
+export class Migration20241119101706 extends Migration {
+    async up(): Promise<void> {
+        this.addSql('alter table "BrevoConfig" add column "folderId" int not null default 1;');
```
RainbowBunchie, line 5: You need to drop the default afterwards again. Generally you should not use default values in the database. This is just useful to fill not nullable fields.

packages/api/src/brevo-config/dto/brevo-config.input.ts
```diff
@@ -14,6 +14,11 @@ export class BrevoConfigInput {
     @IsString()
     @Field()
     senderName: string;
+
+    @IsNotEmpty()
```
RainbowBunchie, line 18:
```suggestion
```

demo/api/schema.gql
```diff
@@ -460,6 +460,7 @@ type BrevoConfig implements DocumentInterface {
   updatedAt: DateTime!
   senderMail: String!
   senderName: String!
+  folderId: Int
```
RainbowBunchie, line 463: This is not nullable right? Why is this not marked with "!" ? 🤔 did the schema not update?

juliawegmayr: b82acde06742ba4bf10ce7aa05202c5a4db90ab6

packages/api/src/mikro-orm/migrations/Migration20241119101706.ts
juliawegmayr: 515452df8cc7b6908f6e5f5d6966959774151771

juliawegmayr: f6a6660efe183ff2248fd57ae5198e590cae382b

RainbowBunchie (approved):

raphaelblum (approved):

PR #233, opened 2024-11-28
add folderId to environment variables

juliawegmayr:
## Description

Add folderId to environment variables to allow overriding default value 1.

## Changeset

https://vivid-planet.atlassian.net/browse/COM-1280


demo/api/src/config/environment-variables.ts
```diff
@@ -142,4 +142,8 @@ export class EnvironmentVariables {
 
     @IsString()
     CAMPAIGN_BASIC_AUTH_PASSWORD: string;
+
+    @IsNumber()
+    @Type(() => Number)
+    BREVO_FOLDER_ID: number;
```
RainbowBunchie, line 148: This should be optional. Else this is a breaking change. Also making it required would not allow overwriting the default but would just be a required config and the default value would never be used

juliawegmayr: Fixed here: 65030841040a19cfdb787abdc949c32f8e0eea01

.env
```diff
@@ -85,3 +85,4 @@ NEXT_PUBLIC_CAMPAIGN_IS_PREVIEW=false
 REDIRECT_URL_FOR_IMPORT=$SITE_URL
 BREVO_ALLOWED_REDIRECT_URL=http://${DEV_DOMAIN:-localhost}${WORKAROUND_DOTENV_ISSUE}:${SITE_PORT}
 CAMPAIGNS_FRONTEND_URL=http://${DEV_DOMAIN:-localhost}${WORKAROUND_DOTENV_ISSUE}:${SITE_PORT} # doesn't work, TODO: add actual mailing frontend
+BREVO_FOLDER_ID=
```
raphaelblum, line 88: I would either set it to 1 here or not set it at all

packages/api/src/config/brevo-module.config.ts
```diff
@@ -15,6 +15,7 @@ export interface BrevoModuleConfig {
             };
             allowedRedirectUrl: string;
             redirectUrlForImport: string;
+            folderId: number;
```
raphaelblum, line 18: this is a breaking change. 
I would make it optional and set it to default 1 in the package, not in the demo

packages/api/src/brevo-api/brevo-api-contact.service.ts
```diff
@@ -186,10 +186,12 @@ export class BrevoApiContactsService {
     }
 
     public async createBrevoContactList(title: string, scope: EmailCampaignScopeInterface): Promise<number | undefined> {
+        const folderId = this.config.brevo.resolveConfig(scope).folderId;
+
         try {
             const contactList = {
                 name: title,
-                folderId: 1, // folderId is required, folder #1 is created by default
+                folderId: folderId,
```
raphaelblum, line 194:
```suggestion
                folderId,
```

juliawegmayr: d0e304fbff0cfe5448d7824589b6744db932ed6f

packages/api/src/config/brevo-module.config.ts
juliawegmayr: 288630f670ac01e9738d773c180115c2bdf41a48

.env
juliawegmayr: ae8df102dbdbce0e39aa40ece8ff10f796cef326

RainbowBunchie (approved):

raphaelblum (approved):

PR #249, opened 2024-12-13
Add sendingState filter in EmailCampaignsGrid

juliawegmayr:
## Description

The column `sendingState` in the `EmailCampaignsGrid` should be filterable.

## Screenshots/screencasts

https://github.com/user-attachments/assets/683bde4e-48f2-4fe2-ba19-9401e2cc63a1

## Changeset

https://vivid-planet.atlassian.net/browse/COM-1451

## Further information

As you can see in the screencast, the `is not` filter option is not working correctly. This bug will be fixed in Comet.

### DIFF
--- a/demo/api/schema.gql
+++ b/demo/api/schema.gql
@@ -666,9 +666,16 @@ input EmailCampaignFilter {
   updatedAt: DateFilter
   title: StringFilter
   subject: StringFilter
+  sendingState: SendingStateEnumFilter
   scheduledAt: DateFilter
   and: [EmailCampaignFilter!]
   or: [EmailCampaignFilter!]
+}
+
+input SendingStateEnumFilter {
+  isAnyOf: [SendingState!]
+  equal: SendingState
+  notEqual: SendingState
 }
 
 input EmailCampaignSort {
--- a/packages/admin/src/emailCampaigns/EmailCampaignsGrid.tsx
+++ b/packages/admin/src/emailCampaigns/EmailCampaignsGrid.tsx
@@ -120,6 +120,15 @@ export function EmailCampaignsGrid({
         ...usePersistentColumnState("EmailCampaignsGrid"),
     };
 
+    const sendingStateOptions: { label: string; value: string }[] = [
+        {
+            value: "SENT",
+            label: intl.formatMessage({ id: "cometBrevoModule.emailCampaign.sent", defaultMessage: "Sent" }),
+        },
+        { value: "DRAFT", label: intl.formatMessage({ id: "cometBrevoModule.emailCampaign.draft", defaultMessage: "Draft" }) },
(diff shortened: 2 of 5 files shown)


demo/api/schema.gql
```diff
@@ -666,11 +666,18 @@ input EmailCampaignFilter {
...
   and: [EmailCampaignFilter!]
   or: [EmailCampaignFilter!]
 }
 
+input SendingStateEnumFilter {
+  isAnyOf: [SendingState!]
+  equal: SendingState
+  notEqual: SendingState
+}
+
```
RainbowBunchie, line 680: The schema changes of the `packages/api/schema.gql ` are missing in this PR, they should be here as wenn

juliawegmayr: Added here: da1a0efd845de909905b71b8c3ae5081c8bea6c9 after fixing generate-scheme file in the main branch.

packages/api/src/email-campaign/dto/email-campaign.filter.ts
```diff
@@ -29,6 +34,12 @@ export class EmailCampaignFilter {
     @Type(() => StringFilter)
     subject?: StringFilter;
 
+    @Field(() => SendingStateEnumFilter, { nullable: true })
+    @ValidateNested()
+    @IsOptional()
```
raphaelblum, line 39: Optional:
it's recommended to use IsUndefinable or IsNullable when possible. They are a little bit more precise. In the future please rather use one of them instead of IsOptional.

raphaelblum (approved):

packages/api/src/email-campaign/dto/email-campaign.filter.ts
juliawegmayr: I used it here, as we still use it in the whole project and I thought I'd rather be consistent here. Maybe I go through the whole project once and replace all of them...? 🤔

packages/admin/src/emailCampaigns/EmailCampaignsGrid.tsx
```diff
@@ -144,8 +153,9 @@ export function EmailCampaignsGrid({
             headerName: intl.formatMessage({ id: "cometBrevoModule.emailCampaign.sendingState", defaultMessage: "Sending State" }),
             renderCell: ({ value }) => <SendingStateColumn sendingState={value} />,
             width: 150,
-            filterable: false,
             sortable: false,
+            type: "singleSelect",
+            valueOptions: sendingStateOptions,
```
RainbowBunchie, line 158: Out of scope if of the task but would it make sense to also make this sortable? I personally would expect to have the grid filtered by status -> `Draft > Scheduled > Sent`

RainbowBunchie (approved):

PR #269, opened 2025-01-21
Add check for valid folderId

juliawegmayr:
Add validation for folderId in the BrevoConfig to only allow entering ids, that actually exist.

Task: [COM-1577](https://vivid-planet.atlassian.net/browse/COM-1577)

[COM-1577]: https://vivid-planet.atlassian.net/browse/COM-1577?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/api/src/brevo-api/brevo-api-folders.service.ts
+++ b/packages/api/src/brevo-api/brevo-api-folders.service.ts
@@ -1,0 +1,34 @@
+import * as Brevo from "@getbrevo/brevo";
+import { Inject, Injectable } from "@nestjs/common";
+import { EmailCampaignScopeInterface } from "src/types";
+
+import { BrevoModuleConfig } from "../config/brevo-module.config";
+import { BREVO_MODULE_CONFIG } from "../config/brevo-module.constants";
+import { BrevoApiFolder } from "./dto/brevo-api-folder";
+
+@Injectable()
+export class BrevoApiFoldersService {
+    private readonly contactsApi: Brevo.ContactsApi;
+
+    constructor(@Inject(BREVO_MODULE_CONFIG) private readonly config: BrevoModuleConfig) {
+        this.contactsApi = new Brevo.ContactsApi();
+    }
+
+    public async getFolders(scope: EmailCampaignScopeInterface): Promise<Array<BrevoApiFolder> | undefined> {
+        const apiKey = this.config.brevo.resolveConfig(scope).apiKey;
+        this.contactsApi.setApiKey(Brevo.ContactsApiApiKeys.apiKey, apiKey);
+
+        // Limit set by Brevo
+        const limit = 50;
+
+        const offset = 0;
+
+        const { response, body } = await this.contactsApi.getFolders(limit, offset);
+
+        if (response.statusCode !== 200) {
+            throw new Error("Failed to get folders");
+        }
+
+        return body.folders as BrevoApiFolder[];
+    }
+}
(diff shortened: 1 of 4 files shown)


packages/api/src/brevo-api/brevo-api-folders.service.ts
```diff
@@ -0,0 +1,34 @@
...
+    constructor(@Inject(BREVO_MODULE_CONFIG) private readonly config: BrevoModuleConfig) {
+        this.contactsApi = new Brevo.ContactsApi();
+    }
+
+    public async getFolders(scope: EmailCampaignScopeInterface): Promise<Array<BrevoApiFolder> | undefined> {
+        const apiKey = this.config.brevo.resolveConfig(scope).apiKey;
+        this.contactsApi.setApiKey(Brevo.ContactsApiApiKeys.apiKey, apiKey);
+
+        // Limit set by Brevo
+        const limit = 50;
```
RainbowBunchie, line 22: Can you only have 50 folders in total or is the limit restricted to 50?

juliawegmayr: The limit is restricted to 50 according to the documentation: https://developers.brevo.com/reference/getfolders-1

RainbowBunchie: But is it possible to create more than 50? Could be that the limit is restricted for the endpoint but you could potentially have 100 folders (limit 50, offset 0/50). So you can only get 50 at a time max. 

I highly doubt someone has more than 50 folders, so its an edge case but it should still work.

juliawegmayr: Ahh, sorry - now I get what you mean. 

I adapted the function to get all folders if there are more than 50: [7d3bf9b](https://github.com/vivid-planet/comet-brevo-module/pull/269/commits/7d3bf9b6b4c707625386951aa826ffa8a97a4695) 

And I adapted the resolver to use the function and added an error message as well, if the folderId is not available: [778ef16](https://github.com/vivid-planet/comet-brevo-module/pull/269/commits/778ef16543c666cb1c6f6b63eb0f185606efcfee)

packages/api/src/brevo-config/brevo-config.resolver.ts
```diff
@@ -50,6 +52,17 @@ export function createBrevoConfigResolver({
             return false;
         }
 
+        private async isValidFolderId({ folderId }: { folderId: number }): Promise<boolean> {
+            for await (const folder of this.brevoFolderIdService.getAllBrevoFolders(Scope)) {
+                if (folder.id === folderId) {
+                    return true;
+                } else {
+                    throw new Error("Folder id is not valid. ");
+                }
```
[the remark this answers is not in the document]
RainbowBunchie: Or better in this case: remove the else / error here, since you throw the error in the calling function anyways in line 108 and 143, so this is not needed here. 
Return true if the folder is found. 
Return false if it is not found 

Alsou you could use array.find here instead of the loop, this would be better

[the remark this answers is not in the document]
juliawegmayr: Thanks - got it :) 
I updated the code here: [06e5338](https://github.com/vivid-planet/comet-brevo-module/pull/269/commits/06e5338625e8e0ad80a4607b4dad4385e3a37510)
I didn't use array.find as I have an AsyncGenerator, so I would need to iterate over it first and push it to a new array before I can use find, right?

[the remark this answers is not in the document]
RainbowBunchie: You are right, I missed that!

RainbowBunchie (approved):
LGTM 👍🏻 please wait for @raphaelblum approval

raphaelblum (approved):

PR #274, opened 2025-02-12
Fix scope passed to brevoTestContactsSelectQuery

juliawegmayr:
## Description

Pass scope correctly to `brevoTestContactsSelectQuery`

## Example

## Changeset


.changeset/wet-news-sit.md
```diff
@@ -0,0 +1,5 @@
+---
+"@comet/brevo-admin": minor
```
raphaelblum, line 2: This should be a patch because it is a bugfix

PR #279, opened 2025-02-21
Fix brevo config migrations

juliawegmayr:
Columns for `scope` were added to the `BrevoConfig` table. This lead to bugs in projects, as scopes are configured differently. Therefor the columns `scope_domain` and `scope_language` were dropped and need to be added with a migration in the project instead.

### DIFF
--- a/.changeset/selfish-years-think.md
+++ b/.changeset/selfish-years-think.md
@@ -1,0 +1,7 @@
+---
+"@comet/brevo-api": patch
+---
+
+Remove `scope` from `BrevoConfig` database migration in module
+
+Add database migration for adding individual `scope` columns to `BrevoConfig` instead.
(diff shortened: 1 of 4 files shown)


.changeset/selfish-years-think.md
```diff
@@ -0,0 +1,7 @@
+---
+"@comet/brevo-api": patch
+---
+
+Remove `scope` from `BrevoConfig` database migration in module
+
+Add database migration for adding individual `scope` columns to `BrevoConfig` instead.
```
raphaelblum, line 7:
```suggestion
A custom database migration must be created in the project to add individual scope columns to BrevoConfig.
```

Maybe like that? So it's clear that this is something that needs to be done in the project.

RainbowBunchie (approved):

raphaelblum (approved):

PR #281, opened 2025-03-06
Allow adding contacts without doi manually

juliawegmayr:
## Description

It should be possible to create contacts manually with the contact form in the admin, without sending a double opt-in mail: 

- Add a field "sendDoubleOptIn" (default: true) 
- Add an alert if "sendDoubleOptIn" is set to false and refactor ContactForm
- Set redirectionUrl to nullable, as it is not required, if no DOI is sent
- Refactor api functionality to differentiate between contacts with and without DOI

This PR only handles manually added contacts. Imported contacts will be handled in an upcoming PR.

## Screenshots/screencasts

Screenshot 2025-03-12 at 08 50 54
Screenshot 2025-03-12 at 08 51 02

## Changeset

https://vivid-planet.atlassian.net/browse/XXXL-969

### DIFF
--- a/packages/admin/src/brevoContacts/form/BrevoContactForm.tsx
+++ b/packages/admin/src/brevoContacts/form/BrevoContactForm.tsx
@@ -1,6 +1,7 @@
 import { DocumentNode, gql, useApolloClient, useQuery } from "@apollo/client";
 import {
     Alert,
+    CheckboxField,
     FinalForm,
     FinalFormSaveButton,
     FinalFormSubmitEvent,
@@ -80,6 +81,7 @@ export function BrevoContactForm({ id, scope, input2State, additionalFormFields,
         let baseInitialValues = {
             email: "",
             redirectionUrl: "",
+            sendDoubleOptIn: true,
         };
 
         if (input2State) {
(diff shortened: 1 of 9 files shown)


packages/admin/src/brevoContacts/form/BrevoContactForm.tsx
```diff
@@ -207,17 +204,49 @@ export function BrevoContactForm({ id, scope, input2State, additionalFormFields,
...
+                                                id="cometBrevoModule.brevoContact.contactAddAlert"
+                                                defaultMessage="The contact will get a double opt-in email to confirm the subscription. After the contact's confirmation, the contact will be added to the corresponding target groups in this scope depending on the contact's attributes. Before the confirmation the contact will not be shown on the contacts page."
+                                            />
+                                        </Alert>
+                                    ) : (
+                                        <Alert severity="error" sx={{ marginBottom: 5 }}>
+                                            <FormattedMessage
+                                                id="cometBrevoModule.brevoContact.contactNoOptInAlert"
+                                                defaultMessage="No Double Opt-In email will be sent. Please ensure that recipients have given their consent before proceeding"
+                                            />
```
raphaelblum, line 233: This should perhaps be written more strongly I think. It should be clarified that the one importing or creating the user is responsible: 

"No Double Opt-In email will be sent. You are responsible for ensuring that recipients have provided their consent before proceeding. If consent has not been given, sending a Double Opt-In email is legally required.

Additionally, the creation of the user will be tracked, and you may need to provide clarification if users report any issues."

packages/api/src/brevo-contact/brevo-contacts.service.ts
```diff
@@ -23,37 +23,42 @@ export class BrevoContactsService {
...
         }
 
-        const created = await this.brevoContactsApiService.createDoubleOptInBrevoContact(
-            { email, redirectionUrl, attributes },
-            brevoIds,
-            templateId,
-            scope,
-        );
+        if (!sendDoubleOptIn) {
+            created = await this.brevoContactsApiService.createBrevoContactWithoutDoubleOptIn({ email, attributes }, brevoIds, scope);
```
raphaelblum, line 53: why is here no templateId?

packages/admin/src/brevoContacts/form/BrevoContactForm.tsx
juliawegmayr: Good point - thanks: bf5b2b4245160683ffd126ee7c9734a859d1adb3

packages/api/src/brevo-contact/brevo-contacts.service.ts
juliawegmayr: Added here: e88734ffbb8336e25939ac50cfc4a3460b714b09

raphaelblum (approved):

PR #282, opened 2025-03-11
Fix CSV Import

juliawegmayr:
## Description

Fix csv import of contacts by using correct types. 

## Example

## Changeset

### DIFF
--- a/.changeset/stale-wolves-type.md
+++ b/.changeset/stale-wolves-type.md
@@ -1,0 +1,5 @@
+---
+"@comet/brevo-api": patch
+---
+
+Fix CSV import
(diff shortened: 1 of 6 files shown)


.changeset/stale-wolves-type.md
```diff
@@ -0,0 +1,5 @@
+---
+"@comet/brevo-api": patch
+---
+
+Fix CSV import
```
raphaelblum, line 5: I would describe it a bit more.
Explaining that they type string[][] was wrong but seemed to worked earlier but with recent updated this lead to an error and you changed it to a JSONObject

raphaelblum (approved):

PR #286, opened 2025-03-14
Change renovate baseBranch to main

thomasdax98:
We should try to make as many updates as possible in main. So I switched the baseBranch for renovate from next to main


raphaelblum:
Thank you @thomasdax98 
still had that on my todo list but couldn't find time for it yet

raphaelblum (approved):

PR #287, opened 2025-03-17
Add logging for created contacts (with form in admin)

juliawegmayr:
## Description

Add logging for contacts, who were created without accepting a double opt-in email via the form in admin. Data is stored in the database only. 

## Changeset

- [x] I have verified if my change requires a changeset

https://vivid-planet.atlassian.net/browse/XXXL-970


packages/api/src/brevo-contact-log/entity/brevo-contact-log.entity.factory.ts
```diff
@@ -0,0 +1,56 @@
...
+        @Field(() => ID)
+        id: string = v4();
+
+        @Property({ columnType: "text" })
+        @Field()
+        email: string;
+
+        @Property()
+        @Field()
+        userId: string;
```
raphaelblum, line 35: Am i correct here:
 - email is from the user that was created in brevo
 - userId is the user that created the new brevo contact

We should clarify that more. Suggestions:
 - importedEmail/...
 - responsibleUserId/adminUserId/...

packages/api/src/brevo-contact-log/entity/brevo-contact-log.entity.factory.ts
```diff
@@ -0,0 +1,56 @@
...
+    createdAt: Date;
+    updatedAt: Date;
+}
+
+export function createBrevoContactLogEntity({ Scope }: { Scope: EmailCampaignScopeInterface }): Type<BrevoContactLogInterface> {
+    @Entity()
+    @ObjectType({
+        implements: () => [DocumentInterface],
+    })
+    class BrevoContactLog implements BrevoContactLogInterface, DocumentInterface {
```
raphaelblum, line 22: I think the name should be more precise. Suggestion: BrevoEmailImportLog

packages/api/src/brevo-contact-log/entity/brevo-contact-log.entity.factory.ts
```diff
@@ -0,0 +1,56 @@
...
+        [OptionalProps]?: "createdAt" | "updatedAt";
+
+        @PrimaryKey({ columnType: "uuid" })
+        @Field(() => ID)
+        id: string = v4();
+
+        @Property({ columnType: "text" })
+        @Field()
+        email: string;
```
raphaelblum, line 31: Is it okay that we lose the information if the user was imported in bulk (with csv import) or manually created?

We could:
 - Save it as a string array (probably bad because potentially a lot of users can be imported in csv file)
 - Save it in an extra table and make a relation here
 - add a type field ("csvImport"/"manualCreation") and another importId where every user that was in the same csv has the same importId

packages/api/src/brevo-contact-log/brevo-contact-log.service.ts
```diff
@@ -0,0 +1,28 @@
...
+
+import { BrevoContactLogInterface } from "./entity/brevo-contact-log.entity.factory";
+
+@Injectable()
+export class BrevoContactLogService {
+    constructor(
+        @InjectRepository("BrevoContactLog") private readonly repository: EntityRepository<BrevoContactLogInterface>,
+        private readonly entityManager: EntityManager,
+    ) {}
+    public async addContactsToLogs(emails: string[], userId: string, scope: EmailCampaignScopeInterface): Promise<BrevoContactLogInterface[]> {
```
raphaelblum, line 14: Do we need to pass an array here? I can't find the usage

juliawegmayr: bf775ec9cc889a12e732a0f20b47ff709ac18c00

packages/api/src/brevo-contact-log/entity/brevo-contact-log.entity.factory.ts
juliawegmayr: I added a field called contactSource (taking "csvImport"/"manualCreation") and one for the importId as well. The importId is not relevant in this PR, as I only add the logs for manual created contacts here. I will update the upcoming PR for the imports as well. 

bb05f806bca0f82f4e42dee66b6dce9894a691fb

packages/api/src/brevo-contact-log/entity/brevo-contact-log.entity.factory.ts
juliawegmayr: 4a0a9cc4b4d5af4fabaec9446aa81f0ae33f7375

PR #290, opened 2025-03-24
Add logs for blacklisted deleted contacts

juliawegmayr:
## Description

Contacts, that are blacklisted should not be imported again without giving their permission (double opt in). Therefor a table is created, where the encrypted mail addresses of the contacts are saved. Only deleted contacts will be saved in the table (all other blacklisted contacts will be checked by the blacklisted/blocked field).

Open TODOs in upcoming MRs:

- Add check for blacklisted contacts in import/create contacts without doi to avoid importing them again

## Changeset

Task: 971


demo/api/src/config/config.ts
```diff
@@ -69,6 +69,9 @@ export function createConfig(processEnv: NodeJS.ProcessEnv) {
         ecgRtrList: {
             apiKey: envVars.ECG_RTR_LIST_API_KEY,
         },
+        encryption: {
+            encryptionKey: envVars.ENCRYPTION_KEY,
```
raphaelblum, line 73:
```suggestion
            key: envVars.ENCRYPTION_KEY,
```

I would either nest it with encryption.key or just write encryptionKey

packages/api/src/blacklisted-contacts/blacklisted-contacts.service.ts
```diff
@@ -0,0 +1,46 @@
...
+
+    constructor(
+        @InjectRepository("BlacklistedContacts") private readonly repository: EntityRepository<BlacklistedContactsInterface>,
+        @Inject(BREVO_MODULE_CONFIG) private readonly config: BrevoModuleConfig,
+        private readonly entityManager: EntityManager,
+    ) {
+        this.secretKey = this.config.encryption.encryptionKey;
+        if (!this.secretKey) {
+            throw new Error("There is no `encryptionKey` defined in the environment variables.");
+        }
```
raphaelblum, line 23: is this check necessary? shouldn't it be required?

.changeset/purple-carrots-bow.md
```diff
@@ -0,0 +1,49 @@
+---
+"@comet/brevo-api": major
+---
+
+Add `BlacklistedContacts` table to store encrypted blacklisted contacts
+
+Use `createBlacklistedContactsEntity` for creating `blacklisted-contacts` entity. Pass `Scope` and add it to the `AppModule`:
```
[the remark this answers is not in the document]
juliawegmayr: efab8f7eaac3a5b282bf25dee1c311268bd38935

packages/api/src/blacklisted-contacts/blacklisted-contacts.service.ts
juliawegmayr: e63c04c40cfd77c4b399c8c08ce71fb02dc6a451

raphaelblum (approved):

PR #283, opened 2025-03-12
Import contacts without doi

juliawegmayr:
## Example

[ ] I have verified if my change requires an example

## Changeset

[ ] I have verified if my change requires a changeset

PR #291, opened 2025-03-24
Check for blacklisted contacts when adding contacts

juliawegmayr:
## Description

Check if a contact is blacklisted/blocked, or already deleted before importing contacts without a doi. Also check, when creating contacts with the form in admin. 

Users should only be able to be added after being blacklisted, if they confirm a DOI.

## Changeset

https://vivid-planet.atlassian.net/browse/XXXL-971

## Further information

- [x] Merge: https://github.com/vivid-planet/comet-brevo-module/pull/281
- [x] Merge: https://github.com/vivid-planet/comet-brevo-module/pull/283
- [x] Merge: https://github.com/vivid-planet/comet-brevo-module/pull/290


packages/api/src/brevo-contact/brevo-contacts.service.ts
```diff
@@ -43,14 +49,20 @@ export class BrevoContactsService {
...
+
+        let created = false;
 
         if (listIds) {
             brevoIds.push(...listIds);
         }
 
         if (!sendDoubleOptIn) {
-            created = await this.brevoContactsApiService.createBrevoContactWithoutDoubleOptIn({ email, attributes }, brevoIds, templateId, scope);
+            const encryptedEmail = encrypt(email, this.secretKey);
```
raphaelblum, line 60: the encrypt function is actually named incorrectly. encryption is reversible and in this case it is a hash, can you please rename it to hash?

raphaelblum (approved):

packages/api/src/brevo-contact/brevo-contacts.service.ts
juliawegmayr: I will do it in another PR (I already opened one, so that I won't forget)

PR #297, opened 2025-04-04
Set import contacts without doi optional

juliawegmayr:
Adding contacts without double opt-in should be an optional feature. 

- [x] Add environment variable to define if it is possible to add contacts without doi
- [x] Use environment variable in the BrevoConfigProvider to adapt the admin frontend
- [x] Set "emailHashKey" variable to optional, as it is only needed, if adding contacts without doi is allowed
- [x] Add checks for emailHashKey in api
- [x] Add checks for allowAddingContactsWithoutDoi to all feature related api functions, that are not already checked by sendDoubleOptIn variable
- [x] Adapt modules 

Tested by creating a test branch, removing the feature from the demo. 

Open to do:
- [x] Correct changesets and changelog

Depends on: 

- [x] https://github.com/vivid-planet/comet-brevo-module/pull/295


demo/admin/src/config.ts
```diff
@@ -25,6 +28,7 @@ export function createConfig() {
         buildNumber: environmentVariables.BUILD_NUMBER,
         commitSha: environmentVariables.COMMIT_SHA,
         campaignUrl: environmentVariables.CAMPAIGN_URL,
+        allowAddingContactsWithoutDoi: toBoolean(environmentVariables.ALLOW_ADDING_CONTACTS_WITHOUT_DOI),
```
juliawegmayr, line 31: @raphaelblum Is there a better way here? The environment variables are all passed as strings, and I wasn't sure how to get a boolean value here.

raphaelblum: you can transform it like here in the environment variables:
https://github.com/vivid-planet/comet-brevo-module/blob/3de2511f238550e8928a7d062c9b03a808e78ede/demo/api/src/config/environment-variables.ts#L16

raphaelblum: ah sorry, yes since it's in the admin it has to be done here in the config.

raphaelblum (approved):

PR #311, opened 2025-05-16
Update to node 22

raphaelblum:
# Description
Update to node 22.
I also had to update graphql-codegen cli
Screenshot 2025-05-16 at 14 54 41

Task: https://vivid-planet.atlassian.net/browse/COM-2017


.nvmrc
```diff
@@ -1 +1 @@
-18
+22.14.0
```
RainbowBunchie, line 1: Is there a reson why you set it to this specific version? Shouldn't 22 be enough here

RainbowBunchie: https://github.com/vivid-planet/comet/blob/main/.nvmrc

raphaelblum: https://github.com/vivid-planet/comet-brevo-module/pull/246/files

I did it because renovate did it.
I used the changes from renovate and added the pipeline changes as well, wasn't sure if I can just push in the renovate PR without breaking anything.

When i set it to 22, renovate will probably change it in the next PR again. I'll investigate if there are other settings in comet

RainbowBunchie (approved):

PR #340, opened 2025-08-07
Update eslint to v9

juliawegmayr:
Comet v8 Preparation: Update eslint and prettier

See: https://docs.comet-dxp.com/docs/migration-guide/migration-from-v7-to-v8/#step-4-update-eslint-and-prettier-pr-4

 Depends on: https://github.com/vivid-planet/comet-brevo-module/pull/339


packages/admin/src/temp/CrudMoreActionsMenu.tsx
```diff
@@ -1,26 +1,26 @@
 // temporary copy from https://github.com/vivid-planet/comet/pull/2115/files
 // remove as soon as it's available in COMET
```
raphaelblum, line 2: can you check in the next PR's if this file is still necessary please?

raphaelblum (approved):

packages/admin/src/temp/CrudMoreActionsMenu.tsx
juliawegmayr: sure 👍

PR #341, opened 2025-08-08
Update to comet v8 - Package Api

juliawegmayr:
## Example

[ ] I have verified if my change requires an example

## Changeset


packages/api/package.json
```diff
@@ -24,32 +24,40 @@
         "lint:eslint": "eslint --max-warnings 0 src/ package.json",
         "lint:prettier": "npx prettier --check './**/*.{js,json,md,yml,yaml}'",
         "lint:tsc": "tsc --noEmit",
+        "mikro-orm": "dotenv -e .env.secrets -e .env.local -e .env -e .env.site-configs -- mikro-orm",
         "test": "jest",
         "test:watch": "jest --watch"
     },
     "dependencies": {
+        "@apollo/server": "^4.0.0",
```
raphaelblum, line 32: why is this needed? is this a new peerDependency, if yes from which package?

PR #346, opened 2025-08-14
Fix comet v8

juliawegmayr:
## Example

[ ] I have verified if my change requires an example

## Changeset

### DIFF
--- a/demo/site/package.json
+++ b/demo/site/package.json
@@ -18,9 +18,8 @@
         "gql:watch": "graphql-codegen --watch",
         "lint": "run-p gql:types generate-block-types && run-p lint:prettier lint:eslint lint:tsc",
         "lint:eslint": "eslint --max-warnings 0 src/ **/*.json --no-warn-ignored",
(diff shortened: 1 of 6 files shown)


demo/site/package.json
```diff
@@ -16,7 +16,7 @@
         "generate-block-types:watch": "chokidar -s \"**/block-meta.json\" -c \"npm run generate-block-types\"",
         "gql:types": "graphql-codegen",
         "gql:watch": "graphql-codegen --watch",
-        "lint": "run-p gql:types generate-block-types && run-p lint:prettier lint:eslint lint:tsc",
+        "lint": "run-s && run-p gql:types generate-block-types && run-p lint:prettier lint:eslint lint:tsc",
```
manuelblum: the empty run-s makes no sense and leads to following error when executing:

Image

```suggestion
        "lint": "run-p gql:types generate-block-types && run-p lint:prettier lint:eslint lint:tsc",
```

